PR #10, opened 2024-09-12
[RSDK-8761] - Use ISO datetime format

seanavery:
## Description

Currently, I am using a non standard datetime format `YYYY-MM-DD_HH-MM-SS` which I adopted from FFmpeg segment examples.

To make the DoCommand endpoint easier to use, this PR moves to standard `ISO 8601` format `YYYY-MM-DDTHH:MM:SS`.

### golang:
- golang does not support standard `ISO 8601`, rather, it uses `RFC 3339` which includes the timezone.
  - It is not possible for the `strftime` formatter in FFmpeg to match this.

## Testing
- storage ✅ 
- save ✅ 
- fetch ✅

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,18 +66,19 @@ The `From` and `To` timestamps are used to specify the start and end times for v
 
 #### Datetime Format
 
-The datetime format used is: `YYYY-MM-DD_HH-MM-SS`
+The datetime format used is ISO 8601: `YYYY-MM-DDTHH:MM:SS`
 
 - `YYYY`: Year (e.g., 2023)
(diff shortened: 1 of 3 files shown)


hexbabe:
> golang does not support standard ISO 8601, rather, it uses RFC 3339 which includes the timezone.
It is not possible for the strftime formatter in FFmpeg to match this.

This is the reason this was rejected right @seanavery 

Any other context I should gather?

seanavery:
@hexbabe Here is a bug report about not supporting UTC
https://trac.ffmpeg.org/ticket/8415

PR #18, opened 2024-10-03
[RSDK-8903] - Implement simple keyframe force interval

seanavery:
## Description

There was an issue where the segmenter would flake out after some time. This was due to a keyframe not being correctly inserted into the beginning of the segment as expected. It turns out, the `gop_size` encoder settings are more of a recommendation than an absolutely firm keyframe interval. Dependenging on scene changes, a keyframe could be inserted earlier than expected throwing off the synchronization between the encoder and segmenter.

FFmpeg codec options actually include a `force_key_frames` expression handler, so we do not even have to hand roll the key frame forces in our image loop! Basically a one line opt call to fix this : )

## Test
- Long running test with `viamrtsp` camera of size 960x480 and 20FPS
  - ~5 hours, ~3000 segments 
  - All playable so far...

https://github.com/user-attachments/assets/e18f8536-995b-4daf-bf58-d447f56ae407

`ffprobe -v error -select_streams v:0 -show_frames -print_format json 2024-10-03_17-48-40.mp4 > frames_info.json`
[frames_info.json](https://github.com/user-attachments/files/17250519/frames_info.json)

Update here after overnight test: ✅ 
- 10 GB storage,  8000 segments, 12 hours runtime

### DIFF
--- a/cam/cam.go
+++ b/cam/cam.go
@@ -31,7 +31,7 @@ const (
 	defaultVideoFormat    = "mp4"
 	defaultUploadPath     = ".viam/capture/video-upload"
 	defaultStoragePath    = ".viam/video-storage"
-	defaultLogLevel       = "info"
+	defaultLogLevel       = "error"
 
 	maxGRPCSize     = 1024 * 1024 * 32 // bytes
 	deleterInterval = 10               // minutes
--- a/examples/save_client.go
+++ b/examples/save_client.go
@@ -9,6 +9,7 @@ package main
 
 import (
 	"context"
+	"math/rand"
 	"os"
 	"time"
 
@@ -17,7 +18,6 @@ import (
 	"go.viam.com/rdk/logging"
 	"go.viam.com/rdk/robot/client"
 	"go.viam.com/utils/rpc"
-	"golang.org/x/exp/rand"
 )
 
 func main() {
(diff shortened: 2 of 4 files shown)


hexbabe (approved):
Glad the fix could be done declaratively as a flag!

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	go.viam.com/rdk v0.40.0
 	go.viam.com/test v1.1.1-0.20220913152726-5da9916c08a2
 	go.viam.com/utils v0.1.98
+	golang.org/x/exp v0.0.0-20240103183307-be819d1f06fc
```
hexbabe, line 17: why this change?

cam/cam.go
```diff
@@ -31,7 +31,7 @@ const (
 	defaultVideoFormat    = "mp4"
 	defaultUploadPath     = ".viam/capture/video-upload"
 	defaultStoragePath    = ".viam/video-storage"
-	defaultLogLevel       = "info"
+	defaultLogLevel       = "error"
```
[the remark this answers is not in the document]
seanavery: I have been meaning to turn down the log level. Mostly, so that it no longer logs every time a new segment is created.

go.mod
seanavery: Forgot to run go mod tidy from last PR.

But thanks for pointing this out, for some reason I was using the external package instead of math/rand.

examples/save_client.go
```diff
@@ -9,6 +9,7 @@ package main
 
 import (
 	"context"
+	"math/rand"
```
seanavery, line 12: Small fix for mod tidy.

go.mod
seanavery: Removed that dep from example script.

PR #19, opened 2024-10-04
[RSDK-8891] - Remove malformed filenames from concat operation

seanavery:
## Description
 
 Simple pr to remove erroneous files from file lookup and storage matcher for concats.
 
 For example  `.DS_STORE` files on macos will now be ignored.
 
 ## Test
 
 - manually tested with .DS_STORE file ✅ 
 - added test video file with incorrect file name ✅


randhid (approved):
thanks for adding a test file, I don't see any changes to the tests but assuming it's being tests somehow?

PR #20, opened 2024-10-07
Manually force keyframes in encoder loop

seanavery:
### Description

Despite having successful 24 hr runs on linux, I am still getting reports of flaky behavior on Darwin. This PR removes FFmpeg handling of keyframe forces entirely, and instead adds in manual keyframe forces in the encoder loop.

- Removes `gop_size` encoder setter entirely to avoid any conflicting behavior.
- Removes keyframe force expression.
- Adds manual keyframe flag every second.

## Testing

- linux ✅ 
  - 3 hour test, 500 segments ✅ 
  - overnight test ✅
- darwin ✅ 
  - 3 hour test, 500 segments ✅ 
  - overnight test ✅


randhid (approved):

hexbabe (approved):
Got it thanks for update

PR #23, opened 2024-10-21
[RSDK-9086] Move from git lfs to artifact

seanavery:
## Description

Git LFS costs money for bandwidth! We have an in-house artifact system for large files. This PR moves tests over to goutils artifact.


hexbabe:
This is cool! I'll review soon

randhid (approved):

Makefile
```diff
@@ -45,6 +45,7 @@ endif
 CGO_CFLAGS := -I$(FFMPEG_BUILD)/include
 GOFLAGS := -buildvcs=false
 export PKG_CONFIG_PATH=$(FFMPEG_BUILD)/lib/pkgconfig
+export PATH := $(PATH):$(shell go env GOPATH)/bin
```
hexbabe, line 48: What are we using `PATH` for here?

tests/config_test.go
```diff
@@ -57,13 +58,13 @@ func cleanVideoStoreDir() error {
 	if err != nil {
 		return err
 	}
-	videoStoreDir := filepath.Join(currentDir, "video-storage")
+	videoStoreDir := filepath.Join(currentDir, "../.artifact/data")
```
hexbabe, line 61: Should we use `artifactStoragePath`

hexbabe (changes requested):

Makefile
seanavery: Just so that bins in go path are available (for `artifact` calls).

tests/config_test.go
seanavery: Ty. Changed.

hexbabe (approved):
Thanks!!

PR #24, opened 2024-12-02
[RSDK-9225] - Separate process frame routines

seanavery:
## Description

In order to guarantee the segmenter gets fed at the desired framerate, we separate frame fetching `fetchFrames` from the camera and the encoding/segmenting step `processFrames`. This guarantees that the time between clips will be exactly the specified `segment_seconds`.

> See [ticket](https://viam.atlassian.net/browse/RSDK-9225) for details about the resulting fetch/save duration mismatch that was downstream of this problem.

## Tests
- Example small resolution 704x480 with viamrtsp that was processing ahead of schedule previously now keeps cadence.
  - Notice the exact 10 second duration between clips.
  - Clips are still playable and look ok.
Screenshot 2024-12-02 at 11 58 31 AM

https://github.com/user-attachments/assets/5a30d199-3da9-4b19-bbac-877411d678bf

### DIFF
--- a/cam/cam.go
+++ b/cam/cam.go
@@ -6,7 +6,9 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"image"
 	"path/filepath"
+	"sync/atomic"
 	"time"
 
 	"go.viam.com/rdk/components/camera"
@@ -47,10 +49,10 @@ type videostore struct {
 	conf   *Config
 	logger logging.Logger
 
-	cam    camera.Camera
-	stream gostream.VideoStream
-
-	workers *utils.StoppableWorkers
+	cam         camera.Camera
+	stream      gostream.VideoStream
+	latestFrame atomic.Pointer[image.Image]
+	workers     *utils.StoppableWorkers
 
 	enc  *encoder
 	seg  *segmenter
@@ -248,7 +250,7 @@ func newvideostore(
 	}
 
 	// Start workers to process frames and clean up storage.
-	vs.workers = utils.NewBackgroundStoppableWorkers(vs.processFrames, vs.deleter)
+	vs.workers = utils.NewBackgroundStoppableWorkers(vs.fetchFrames, vs.processFrames, vs.deleter)
 
 	return vs, nil
 }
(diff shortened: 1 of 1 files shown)


hexbabe (commented):
This makes sense to me. Essentially instead of relying on `GetImage` to keep its frequency constant, we force a frequency by reading the latest frame consistently, while updating latest frame. A question I have is: where is `processFrames` being called now? Doesn't show in the diff. Would be helpful to see to assess how contention for the atomic pointer will affect things.

hexbabe:
Btw the video clip you attached in the description isn't playable

seanavery:
> his makes sense to me. Essentially instead of relying on GetImage to keep its frequency constant, we force a frequency by reading the latest frame consistently, while updating latest frame.
Yep exactly -- thats the idea. If the GetImage loop cannot keep up, we will process duplicate frames which isnt that big of a deal.

> A question I have is: where is processFrames being called now? Doesn't show in the diff. Would be helpful to see to assess how contention for the atomic pointer may affect things.
processFrames is just a background routine that gets spawned by stoppable workers (`NewBackgroundStoppableWorkers`), it is not explicitly called.

> 
Btw the video clip you attached in the description isn't playable for me

hm... Maybe github bug? Just tried on another browser and seems to be playing ok.

randhid:
I can see video on firefox. Reading pr now.

randhid (approved):

cam/cam.go
```diff
@@ -337,10 +339,10 @@ func (vs *videostore) Properties(_ context.Context) (camera.Properties, error) {
...
 }
 
-// processFrames reads frames from the camera, encodes, and writes to the segmenter
-// which chunks video stream into clip files inside the storage directory.
-// TODO(seanp): Should this be throttled to a certain FPS?
-func (vs *videostore) processFrames(ctx context.Context) {
+// fetchFrames reads frames from the camera and stores the decoded image
+// in the latestFrame atomic pointer. This routine runs as fast as possible
+// to keep the latest frame up to date.
+func (vs *videostore) fetchFrames(ctx context.Context) {
```
randhid, line 345: We may want a benchmark tests for this, either now or a future pr.

seanavery: sounds good -- have performance ticket up next so include benchmarks in that diff.

PR #25, opened 2024-12-03
[RSDK-9420] - Update to image

seanavery:
## Description

This updates to use new `Image` endpoint instead of `stream.Next` when querying from the source camera. By removing an extra JPEG encode/decode step we are seeing big performance improvements.

Lots of extra diff from linter upgrade, only core change is removing stream and adding Image call.

## Tests
- tested performance improvement with large frame size (960x1080)
  - before: `cam/cam.go:358 stream.Next took 81.108933ms`
  - after: `cam/cam.go:374  DecodeImageFromCamera took 26.321174ms`
- manually tested segments look ok and save cmds still working


cam/cam.go
```diff
@@ -50,7 +51,6 @@ type videostore struct {
 	logger logging.Logger
 
 	cam         camera.Camera
-	stream      gostream.VideoStream
```
randhid, line 53: 🥳 🍾 🍻 🔥 🎆 🎉

cam/concater.go
```diff
@@ -138,13 +138,13 @@ func (c *concater) concat(from, to time.Time, path string) error {
 
 	// Copy codec info from input to output context. This is necessary to ensure
 	// we do not decode and re-encode the video data.
-	for i := 0; i < int(inputCtx.nb_streams); i++ {
+	for i := range make([]int, int(inputCtx.nb_streams)) {
```
randhid, line 141: I think this can be 
```suggestion
	for i := range int(inputCtx.nb_streams) {
```

cam/encoder.go
```diff
@@ -189,13 +193,13 @@ func imageToYUV422(img image.Image) ([]byte, error) {
...
 	ySize := width * height
-	halfWidth := width / 2
+	halfWidth := width / subsampleFactor
 	uSize := halfWidth * height
 	vSize := uSize
 
 	rawYUV := make([]byte, ySize+uSize+vSize)
 
-	for y := 0; y < height; y++ {
+	for y := range make([]int, height) {
```
randhid, line 202: same comment as `concater.go:141`

cam/segmenter.go
```diff
@@ -83,7 +85,7 @@ func newSegmenter(
 		return nil, fmt.Errorf("failed to copy codec parameters %s", ffmpegError(ret))
 	}
 
-	segmentLengthCStr := C.CString(fmt.Sprintf("%d", clipLength))
+	segmentLengthCStr := C.CString(strconv.Itoa(clipLength))
```
randhid, line 88: This is tested to prove it was equivalent to the behaviour before?

randhid (approved):
LGTM! Few nits here and there.

Makefile
```diff
@@ -81,7 +81,7 @@ tool-install:
 
 lint: tool-install $(FFMPEG_BUILD)
 	go mod tidy
-	CGO_CFLAGS=$(CGO_CFLAGS) GOFLAGS=$(GOFLAGS) $(TOOL_BIN)/golangci-lint run -v --fix --config=./etc/.golangci.yaml
+	CGO_CFLAGS=$(CGO_CFLAGS) GOFLAGS=$(GOFLAGS) $(TOOL_BIN)/golangci-lint run -v --fix --config=./etc/.golangci.yaml --timeout=2m
```
hexbabe, line 84: curious were you running into hangs on lint?

hexbabe (approved):

cam/concater.go
seanavery: ah, thats a lot better -- will change range iterators to this pattern.

cam/segmenter.go
seanavery: Yep, segment clip lengths are the correct size.

Makefile
seanavery: Yeah not sure why it is taking so much longer now...

https://github.com/viam-modules/video-store/actions/runs/12162796386/job/33920591465

cam/utils.go
```diff
@@ -78,8 +78,9 @@ func lookupCodecTypeByID(codecID C.enum_AVCodecID) codecType {
 
 // ffmpegError returns a string representation of the ffmpeg error code.
 func ffmpegError(ret C.int) string {
-	var errbuf [256]C.char
-	C.av_strerror(ret, &errbuf[0], 256)
+	const errbufSize = 256
```
randhid, line 81: Ah one nit - these never change so can allocate outside the function call - next pr!

PR #28, opened 2024-12-05
post-image profiling

hexbabe:


seanavery:
@hexbabe Could builds with pprof insertions be made into a Makefile script?

hexbabe:
> @hexbabe Could builds with pprof insertions be made into a Makefile script?

95% sure it's possible

PR #36, opened 2025-01-31
[RSDK-9879] - Make datamanger sync dep implcit

seanavery:
## Description
When I first wrote video-store, I did not understand implicit dependencies, so just fixing this now. Also, a user recently wanted to use the module without a sync service so wanted to make that possible.

- Gets rid of hacky explicit datamanager dep check.
- Supports implicit datamanager dep.
- Makes sync attribute optional.

## Tests

- Added automated config tests ✅ 
- Manual tests:
  - Implicit sync dep succeeds ✅ 
  - Missing sync attr succeeds ✅ 
  - Mis-named sync dep fails ✅ 
  - Sync attr with no service fails ✅


hexbabe (approved):
Code LGTM 

Also what Rand says makes sense for extra carefulness

PR #37, opened 2025-02-06
Raw segmenter

seanavery:
## Description

## Todo:

### DIFF
--- a/cam/rawsegmenter.go
+++ b/cam/rawsegmenter.go
@@ -1,0 +1,281 @@
+package videostore
+
+/*
+#cgo pkg-config: libavutil libavcodec libavformat libswscale
+#include <libavfilter/avfilter.h>
+#include <libavutil/avutil.h>
+#include <libavformat/avformat.h>
+#include <libavutil/dict.h>
+#include <libavutil/error.h>
+#include <libavcodec/avcodec.h>
+*/
+import "C"
+
+import (
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"unsafe"
+
+	"go.viam.com/rdk/logging"
+)
+
+type RawSegmenter struct {
+	logger      logging.Logger
+	storagePath string
+	outCtx      *C.AVFormatContext
+}
+
+func NewRawSegmenter(
+	logger logging.Logger,
+	storagePath string,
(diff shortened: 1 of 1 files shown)


cam/rawsegmenter.go
```diff
@@ -0,0 +1,254 @@
...
+		return fmt.Errorf("failed to set strftime: %s", ffmpegError(ret))
+	}
+
+	// Open the output file for writing
+	ret = C.avformat_write_header(fmtCtx, &opts)
+	if ret < 0 {
+		return fmt.Errorf("failed to write header: %s", ffmpegError(ret))
+	}
+
+	// Need to reset stream time base to 1/1000
```
seanavery, line 148: [seanp] Do we stil need this hack to reset time base after writing header?

cam/rawsegmenter.go
```diff
@@ -0,0 +1,254 @@
...
+	C.memcpy(unsafe.Pointer(avpkt.data), unsafe.Pointer(&payload[0]), C.size_t(len(payload)))
+	avpkt.size = C.int(len(payload))
+	if avpkt.size == 0 {
+		return errors.New("empty packet")
+	}
+	if avpkt.data == nil {
+		return errors.New("nil packet data")
+	}
+	avpkt.pts = C.int64_t(pts)
+	avpkt.dts = C.int64_t(pts)
```
seanavery, line 178: [seanp] Do we need to set the DTS (decode-timstamp)?

cam/rawsegmenter.go
```diff
@@ -0,0 +1,254 @@
...
+	ret := C.av_interleaved_write_frame(rs.outCtx, avpkt)
+	if ret < 0 {
+		return fmt.Errorf("failed to write frame: %s", ffmpegError(ret))
+	}
+	return nil
+}
+
+// Close closes the segmenter and writes the trailer to prevent corruption
+// when exiting early in the middle of a segment.
+func (rs *RawSegmenter) Close() {
```
seanavery, line 194: [seanp] This handles the case when the program early exits and we need to terminate a clip before it is finished.

nicksanford:
moved https://github.com/viam-modules/video-store/pull/39

PR #39, opened 2025-02-11
Rsdk 9919 move video store logic into viamrtsp

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-9919)

Derived from https://github.com/viam-modules/video-store/pull/37

## Description

## Todo:

### DIFF
--- a/cam/cam.go
+++ b/cam/cam.go
@@ -2,6 +2,7 @@
 package videostore
 
 /*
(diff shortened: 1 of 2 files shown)


nicksanford:
moved https://github.com/viam-modules/video-store/pull/40

PR #40, opened 2025-02-11
Rsdk 9919 move video store logic into viamrtsp

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-9919)

Derived from https://github.com/viam-modules/video-store/pull/37

## Description

## Todo:


videostore/rawsegmenter.go
```diff
@@ -0,0 +1,286 @@
...
+	// Maybe we can get this from ONVIF or the RTSP stream SDP.
+	codecCtx.width = C.int(hsps.Width())
+	codecCtx.height = C.int(hsps.Height())
+	// codecCtx.pix_fmt = C.AV_PIX_FMT_YUV420P
+	// codecCtx.time_base.num = 1
+	// stream.time_base.den = 1
+
+	// Set up the extradata for the codec context. This allows the muxer to know how to
+	// decode the stream without having to read the SPS/PPS from the stream.
+	extradata, err := buildAVCExtradata(sps, pps)
```
nicksanford, line 107: we still need to test if we need this

videostore/rawsegmenter.go
```diff
@@ -0,0 +1,287 @@
...
+	codecCtx := C.avcodec_alloc_context3(codec)
+	if codecCtx == nil {
+		return errors.New("failed to allocate codec context")
+	}
+
+	// TODO(seanp); Size and pixel format is hardcoded for now. Need to make this dynamic.
+	// Maybe we can get this from ONVIF or the RTSP stream SDP.
+	codecCtx.width = C.int(hsps.Width())
+	codecCtx.height = C.int(hsps.Height())
+	// codecCtx.pix_fmt = C.AV_PIX_FMT_YUV420P
```
seanavery, line 102: Seems like this is working but thinking that the default h264 codec context pix_fmt is `YUV420P`. 

This should be pretty safe since almost all RTSP cameras will use this format but may want to add an argument here for pixel format in the case that we get `yuvj420p` or `yuv422p`.

videostore/rawsegmenter.go
```diff
@@ -0,0 +1,287 @@
...
+	// codecCtx.time_base.num = 1
```
seanavery, line 103: Can we at least add a comment here that we are using the default time base? Might be better to be more explicit here and pass in the desired time base.

seanavery:
Manual Regression Test for Frame Path:

Base Case (h264 480p viamrtsp source camera on an orin-nano)
- Segment files playable with correct duration ✅ 
- Concat operations create playable video (correct playback speed) with segment files ✅ 
- Concat operations create playable video (correct playback speed) with older segment files from stable release ✅ 
- Reconfigure goes through and continues to write segment files ✅ 
- Change in size of camera handled (closes out current segment and continues writing new segments) ✅ 

Test Coverage:
- arm64 (pi5) ✅ 
  - 480p(704x480) ✅
    - h264 ✅
      - segments ✅
      - concat ✅
    - h265 ✅
      - segments ✅
      - concat ✅
   - 2k (2560x1440)
     - h264 ✅
       - segments ✅
       - concat ✅
     - h265 ✅
       - segments ✅
       - concat ✅
   - 4k (3840x2160) ✅ 
     - h265
       - segments ✅ 
       - concat ✅ 
 - x86 (framework laptop) ✅ 
   - 480p (h264)
     - segments ✅ 
     - concat ✅ 
   - 4k (h265)
     - segments ✅ 
     - concat ✅

Makefile
```diff
@@ -42,31 +42,65 @@ FFMPEG_OPTS ?= --prefix=$(FFMPEG_BUILD) \
...
 
-build: $(BIN_OUTPUT_PATH)/video-store
+all: $(FFMPEG_BUILD) $(BIN_OUTPUT_PATH)/video-store $(BIN_OUTPUT_PATH)/concat
 
-build-pprof:
-	$(MAKE) build PPROF_ENABLED=true
+$(BIN_OUTPUT_PATH)/video-store: videostore/*.go cmd/module/*.go $(FFMPEG_BUILD) $(BUILD_TAG_FILE)
+	go mod tidy
+	CGO_LDFLAGS="$(CGO_LDFLAGS)" CGO_CFLAGS="$(CGO_CFLAGS)" go build -tags "$(BUILD_TAGS)" -o $(BIN_OUTPUT_PATH)/video-store cmd/module/cmd.go
+	echo "$(BUILD_TAGS)" > $(BUILD_TAG_FILE)
```
seanavery, line 68: This is from pprof stuff.

seanavery (approved):
Happy with manual test through and in-person walk through of code.

PR #42, opened 2025-02-18
Remove segment seconds from readme

randhid:


hexbabe (approved):
nice thanks for adding the explicit dep note for data manager that one gets me tripped up every time

PR #47, opened 2025-02-26
Add h265 embeded video store support

nicksanford:
1. Add support for saving h265 video via the rawpacketizer including tags for QuickTime compatibility.
2. Change the signature of rawpacketizer to allow the caller to provide dts and not require extradata to be provided prior to writing the first packet.
3. Fix bug where if 2 whole segement files were concatenated the dts would repeat causing concatenation to fail.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ FFMPEG_OPTS ?= --prefix=$(FFMPEG_BUILD) \
                --enable-static \
                --enable-libx264 \
                --enable-decoder=hevc \
-               --enable-decoder=libx264 \
+               --enable-decoder=h264 \
                --enable-gpl \
                --enable-encoder=libx264 \
                --enable-muxer=segment \
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ FFMPEG_OPTS ?= --prefix=$(FFMPEG_BUILD) \
                --disable-everything \
                --enable-static \
                --enable-libx264 \
+               --enable-decoder=hevc \
                --enable-gpl \
                --enable-encoder=libx264 \
                --enable-muxer=segment \
@@ -34,10 +35,12 @@ FFMPEG_OPTS ?= --prefix=$(FFMPEG_BUILD) \
                --enable-demuxer=mov \
                --enable-demuxer=mp4 \
                --enable-parser=h264 \
+               --enable-parser=hevc \
                --enable-protocol=file \
                --enable-protocol=concat \
                --enable-protocol=crypto \
                --enable-bsf=h264_mp4toannexb \
+               --enable-bsf=hevc_mp4toannexb \
                --enable-decoder=mjpeg
 
 GOFLAGS := -buildvcs=false
@@ -51,9 +54,9 @@ ifeq ($(SOURCE_OS),linux)
 	SUBST = -l:libx264.a
 endif
 ifeq ($(SOURCE_OS),darwin)
-	SUBST = $(HOMEBREW_PREFIX)/Cellar/x264/r3108/lib/libx264.a
+	SUBST = $(HOMEBREW_PREFIX)/Cellar/x264/r3108/lib/libx264.a 
 endif
-CGO_LDFLAGS = $(subst -lx264, $(SUBST),$(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) pkg-config --libs $(FFMPEG_LIBS))) 
+CGO_LDFLAGS = $(subst -lx264, $(SUBST), $(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) pkg-config --libs $(FFMPEG_LIBS)))
 export PATH := $(PATH):$(shell go env GOPATH)/bin
 
 .PHONY: lint tool-install test clean clean-all clean-ffmpeg module build valgrind
--- a/Makefile
+++ b/Makefile
@@ -51,12 +51,12 @@ OBJS := $(subst $(SRC_DIR), $(BUILD_DIR), $(SRCS:.c=.o))
 PKG_CONFIG_PATH = $(FFMPEG_BUILD)/lib/pkgconfig
 CGO_CFLAGS = $(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) pkg-config --cflags $(FFMPEG_LIBS))
 ifeq ($(SOURCE_OS),linux)
-	SUBST = -l:libx264.a
+	SUBST = -l:libx264.a 
 endif
 ifeq ($(SOURCE_OS),darwin)
 	SUBST = $(HOMEBREW_PREFIX)/Cellar/x264/r3108/lib/libx264.a 
 endif
-CGO_LDFLAGS = $(subst -lx264, $(SUBST), $(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) pkg-config --libs $(FFMPEG_LIBS)))
+CGO_LDFLAGS = $(subst -lx264, $(SUBST),$(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) pkg-config --libs $(FFMPEG_LIBS))) 
 export PATH := $(PATH):$(shell go env GOPATH)/bin
 
 .PHONY: lint tool-install test clean clean-all clean-ffmpeg module build valgrind
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ FFMPEG_OPTS ?= --prefix=$(FFMPEG_BUILD) \
                --enable-static \
                --enable-libx264 \
                --enable-decoder=hevc \
+               --enable-decoder=libx264 \
                --enable-gpl \
                --enable-encoder=libx264 \
                --enable-muxer=segment \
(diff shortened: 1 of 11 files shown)


videostore/rawsegmenter.c
```diff
@@ -8,47 +8,36 @@
...
 #include <stdio.h>
 #include <string.h>
-int video_store_raw_seg_init_h264(struct raw_seg_h264 **ppRS, // OUT
-                                  const int segmentSeconds,   // IN
-                                  const char *outputPattern,  // IN
-                                  const char *extradata,      // IN
-                                  const size_t extradataSize, // IN
-                                  const int width,            // IN
-                                  const int height            // IN
+int video_store_raw_seg_init(struct raw_seg **ppRS,     // OUT
```
nicksanford, line 11: we no longer need exradata we now expect the callers of `video_store_raw_seg_write_packet` to send what was previously sent in extradata at the beginning of the packet.

videostore/rawsegmenter.c
```diff
@@ -58,71 +47,63 @@ int video_store_raw_seg_init_h264(struct raw_seg_h264 **ppRS, // OUT
...
     av_log(NULL, AV_LOG_ERROR,
-           "video_store_raw_seg_init_h264 failed to copy codec parameters");
+           "video_store_raw_seg_init failed to copy codec parameters\n");
     goto cleanup;
   }
+  if (codec->id == AV_CODEC_ID_H265) {
+    // this is needed  to make h265 videos playable on apple devices
+    // https://trac.ffmpeg.org/wiki/Encode/H.265#FinalCutandApplestuffcompatibility
+    // https://stackoverflow.com/questions/50565912/h265-codec-changes-from-hvc1-to-hev1
+    stream->codecpar->codec_tag = MKTAG('h', 'v', 'c', '1');
```
nicksanford, line 68: I've confirmed that the videos (both h264 andf h265) produced by the concater are playable on mac by vlc and quicktime, and on android

videostore/concat.c
```diff
@@ -74,8 +75,6 @@ int video_store_concat(const char *concat_filepath, const char *output_path) {
              i, av_err2str(ret));
       goto cleanup;
     }
-    // Let ffmpeg handle the codec tag for us.
-    outStream->codecpar->codec_tag = 0;
```
nicksanford, line 78: we need to remove this or the `codec_tag` written by rawsegmenter.c (which is needed for quicktime to be able to play the produced video) will be removed from the concated video.

cmd/raw-segmenter-c/main.c
```diff
@@ -42,12 +44,10 @@ int main(int argc, char *argv[]) {
...
 
-  ret = video_store_raw_seg_init_h264(
-      &rs, 30, "./mp4s/%Y-%m-%d_%H-%M-%S.mp4",
-      sqlite3_column_blob(statement, 0),
-      (size_t)sqlite3_column_bytes(statement, 0), 2560, 1440);
+  ret = video_store_raw_seg_init_h265(&rs, 30, "./mp4s/%Y-%m-%d_%H-%M-%S.mp4",
+                                      2960, 1668);
   if (ret != VIDEO_STORE_RAW_SEG_RESP_OK) {
-    printf("video_store_raw_seg_init_h264 failed: %d\n", ret);
+    printf("video_store_raw_seg_init_h265 failed: %d\n", ret);
```
nicksanford, line 50: this file is now broken, re-valgrind this.

videostore/rawsegmenter.c
```diff
@@ -147,37 +128,69 @@ int video_store_raw_seg_init_h264(struct raw_seg_h264 **ppRS, // OUT
...
+  }
+  return video_store_raw_seg_init(ppRS, segmentSeconds, outputPattern, width,
+                                  height, codec);
+}
+
+int video_store_raw_seg_write_packet(struct raw_seg *rs,       // IN
+                                     const char *payload,      // IN
+                                     const size_t payloadSize, // IN
+                                     const int64_t pts,        // IN
+                                     const int64_t dts,        // IN
```
nicksanford, line 167: we now require the caller to specify dts.

videostore/concat.c
```diff
@@ -129,18 +129,27 @@ int video_store_concat(const char *concat_filepath, const char *output_path) {
         AV_ROUND_NEAR_INF | AV_ROUND_PASS_MINMAX);
     packet->pos = -1;
 
+    if (packet->dts <= prevDts) {
```
nicksanford, line 132: It turns out that when the concatenator goes from processing the last packet of one file to processing the first packet of the next file the dts will be repeated and `AV_PKT_FLAG_DISCARD` won't be set on the pkt->flags. I think `AV_PKT_FLAG_DISCARD` is only set if you are concatenating a subset of a file. From what I can tell if you are concatenating entire files, it won't be set.

Previously in my testing I was only testing 30 second ranges and as a result there were only either 1 whole file or 2 partial files being concatenated, one of which had `AV_PKT_FLAG_DISCARD`  set.

Now we need to explicitly check for repeating dts and skip it.

model/camera/camera.go
```diff
@@ -155,8 +155,9 @@ func (c *component) Properties(_ context.Context) (camera.Properties, error) {
 }
 
 // Close closes the video storage camera component.
-func (c *component) Close(ctx context.Context) error {
-	return c.videostore.Close(ctx)
+func (c *component) Close(_ context.Context) error {
```
nicksanford, line 158: we never needed context, so this reduces boilerplate

videostore/rawsegmenter.c
```diff
@@ -58,71 +47,63 @@ int video_store_raw_seg_init_h264(struct raw_seg_h264 **ppRS, // OUT
...
-    goto cleanup;
-  }
-  codecCtx->extradata_size = (int)extradataSize;
 
-  // memcpy can't fail
-  memcpy(codecCtx->extradata, extradata, extradataSize);
-  /* // Copy the codec parameters from the input stream to the output
-   * stream. This is equivalent */
-  /* // to -c:v copy in ffmpeg cli. This is needed to make sure we do not
-   * re-encode the stream. */
```
nicksanford, line 82: this is no longer needed as the caller will now be expected to prepend sps, pps, and for h265 vps in the WritePacket call

nicksanford:
I manually tested the video-store module produced from this build on a pi 5 & confirmed it still works.

videostore/concat.c
nicksanford: I also confirmed this doensn't result in any hitching or artifacts in the videos produced, both in h264 and h265

videostore/rawsegmenter.c
```diff
@@ -147,37 +128,69 @@ int video_store_raw_seg_init_h264(struct raw_seg_h264 **ppRS, // OUT
...
-                                          const size_t payloadSize, // IN
-                                          const int64_t pts,        // IN
-                                          const int isIdr           // IN
+int video_store_raw_seg_init_h264(struct raw_seg **ppRS,     // OUT
+                                  const int segmentSeconds,  // IN
+                                  const char *outputPattern, // IN
+                                  const int width,           // IN
+                                  const int height           // IN
+) {
+  const struct AVCodec *codec = avcodec_find_decoder(AV_CODEC_ID_H264);
```
nicksanford, line 137: the previous code looked for an encoder... we don't need to do that we can instead look for a decoder (which is semantically what we need).

model/camera/camera.go
```diff
@@ -155,8 +155,9 @@ func (c *component) Properties(_ context.Context) (camera.Properties, error) {
...
+	c.videostore.Close()
```
randhid, line 159: We never error either?

nicksanford: As the code is currently written, correct.

cmd/raw-segmenter-c/main.c
```diff
@@ -9,14 +10,16 @@ int main(int argc, char *argv[]) {
   if (argc != 2) {
     printf("usage: %s <sqlite.db>\n", argv[0]);
     printf("TABLE should have the following schema:\n");
-    printf("CREATE TABLE extradata(id INTEGER NOT NULL PRIMARY KEY, data "
-           "BLOB);\n");
+
+    printf("CREATE TABLE extradata(id INTEGER NOT NULL PRIMARY KEY, isH264 "
```
nicksanford, line 14: I ran this through valgrind using https://github.com/nicksanford/video-vcr to produce the db file.

seanavery (commented):
Done with first pass here -- looking good!

Still need to get more context from the viamrtsp side on how DTS is handled in the gortsplib layer.

videostore/concat.c
nicksanford: 1. They have been equivalent in all the cases I've seen but I saw in the gortsplib examples of writing an ts file to disk they pass dts and pts separately, so I followed that pattern here in case there are cameras out there which will send dts that is different than pts.

https://github.com/bluenviron/gortsplib/blob/main/examples/client-play-format-h265-to-disk/mpegts_muxer.go#L105

videostore/concat.c
```diff
@@ -15,7 +15,8 @@ int video_store_concat(const char *concat_filepath, const char *output_path) {
   int outputPathOpened = 0;
   const AVInputFormat *inputFormat = av_find_input_format("concat");
   if (inputFormat == NULL) {
-    av_log(NULL, AV_LOG_ERROR, "video_store_concat failed to find input format\n");
+    av_log(NULL, AV_LOG_ERROR,
```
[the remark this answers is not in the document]
nicksanford: I think you are right

Makefile
```diff
@@ -25,6 +25,8 @@ FFMPEG_OPTS ?= --prefix=$(FFMPEG_BUILD) \
                --disable-everything \
                --enable-static \
                --enable-libx264 \
+               --enable-decoder=hevc \
+               --enable-decoder=libx264 \
```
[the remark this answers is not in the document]
nicksanford: good call, updated

seanavery (approved):
LGTM.

Manually tested frame path segmenter still works and checked concater looks good.

PR #49, opened 2025-03-04
[RSDK-10011] - Remove stream method

seanavery:
## Description

Simply bumping rdk and removing stream interface method.

## Testing
- Updated GRPC limit fetch test to exceed new 32MB limit ✅ 
- Manually tested video-store frame path still works ✅
- Streaming server does not panic ✅


.artifact/config.json
```diff
@@ -7,5 +7,6 @@
   },
   "ignore": [
     ".DS_Store"
-  ]
+  ],
+  "source_pull_size_limit": 20971520
```
seanavery, line 11: https://github.com/viamrobotics/goutils/blob/main/artifact/config.go#L45

randhid (approved):

PR #52, opened 2025-03-06
[CONSULT-764] - Fix race condition between mime handler and encoder

seanavery:
## Description

See [ticket](https://viam.atlassian.net/browse/CONSULT-764) for log details of occassional segfault in the encoder. This problem is due to a race condition where the mime handler attempts to operate on the `latestFrame` concurrently with the encoder. 

To fix this, we simply add locking around mime and encoder handling of the latestFrame.

## Testing

- Manually tested that frame path video-store is working well with h264 and h265 rtsp cameras ✅ 
- Did not segfault over an overnight 12 hour run ✅ 
- Did not segfault when changing size of source camera 10 times ✅ 
- Segments are still playable and do not seem to be skipping significantly more frames with lock introduced ✅ 

https://github.com/user-attachments/assets/e21557c4-4113-4fd0-92b4-48d2820131de

Having a hard time setting up a reliable way to make the segfault case happen as it requires concurrently calling the mime handler and encoder in separate routines (which is why we are seeing segfaults very infrequently).

### DIFF
--- a/videostore/videostore.go
+++ b/videostore/videostore.go
@@ -11,7 +11,7 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
-	"sync/atomic"
+	"sync"
 	"time"
 
 	"go.viam.com/rdk/components/camera"
@@ -61,8 +61,9 @@ type videostore struct {
 	config Config
 	logger logging.Logger
 
-	latestFrame atomic.Pointer[C.AVFrame]
-	workers     *utils.StoppableWorkers
+	latestFrame   *C.AVFrame
+	latestFrameMu sync.Mutex
+	workers       *utils.StoppableWorkers
 
 	rawSegmenter *rawSegmenter
 	segmenter    *segmenter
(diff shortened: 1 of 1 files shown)


videostore/videostore.go
```diff
@@ -339,27 +340,32 @@ func (vs *videostore) fetchFrames(ctx context.Context, cam camera.Camera) {
...
-					vs.logger.Error("failed to convert yuyv422 to yuv420p", err)
-					continue
-				}
-			case rutils.MimeTypeJPEG, rutils.MimeTypeJPEG + "+" + rutils.MimeTypeSuffixLazy:
-				frame, err = mimeHandler.decodeJPEG(bytes)
-				if err != nil {
-					vs.logger.Error("failed to decode jpeg", err)
-					continue
+			// Transform image bytes to yuv420p based on the mimetype.
+			func() {
```
seanavery, line 344: Wrapping in a function so we can always defer unlocking making it simpler to handle error cases.

videostore/videostore.go
```diff
@@ -376,46 +382,14 @@ func (vs *videostore) processFrames(ctx context.Context, encoder *encoder) {
...
-			// ERROR   video-store-module.modmanager.video-storage.StdErr      pexec/managed_process.go:277
-			// \_ github.com/viam-modules/video-store/videostore._Cfunc_avcodec_send_frame(0x14b80ae50, 0x14b80a9c0)
-			// ERROR   video-store-module.modmanager.video-storage.StdErr      pexec/managed_process.go:277
-			// \_      _cgo_gotypes.go:1270 +0x34 fp=0x14000818cd0 sp=0x14000818ca0 pc=0x105736b44
-			// ERROR   video-store-module.modmanager.video-storage.StdErr      pexec/managed_process.go:277
-			// \_ github.com/viam-modules/video-store/videostore.(*encoder).encode.func1(0x163?, 0x14b80a9c0)
-			// ERROR   video-store-module.modmanager.video-storage.StdErr      pexec/managed_process.go:277
-			// \_      /Users/nicksanford/code/video-store/videostore/encoder.go:178 +0x68 fp=0x14000818d10 sp=0x14000818cd0 pc=0x105738888
-
-			// the encoder and segmenter need to be moved into a single object with a single lifetime and not pass codecCtx
```
seanavery, line 416: I think this comment is off-base here. If you look at the log dump, we see that it fails when sending the frame to the encoder, and not on initialization from size changes.

videostore/videostore.go
```diff
@@ -61,8 +61,9 @@ type videostore struct {
 	config Config
 	logger logging.Logger
 
-	latestFrame atomic.Pointer[C.AVFrame]
-	workers     *utils.StoppableWorkers
+	latestFrame   *C.AVFrame
+	latestFrameMu sync.Mutex
```
seanavery, line 65: Using a normal lock instead of `sync.RWMutex` since ratio of reads to writes is near 1.

seanavery:
@randhid Would be nice to get this diff into a new video-store release sooner rather than later so that we fix segfault issues before video-store embedded into viamrtsp is ready.

videostore/videostore.go
randhid: Add the reasoning to the comment above.

randhid (commented):
LGTM honestly! Nice testing

randhid (approved):

PR #55, opened 2025-03-11
DOCS-3678: Update README

npentrel:
I believe this is correct right?


seanavery (approved):
Looks good! That is the resulting behavior. Thanks

PR #56, opened 2025-03-12
[RSDK-10224] - Delete concated file during fetch

seanavery:
## Description

Simply deleting file in fetch after it is read into go bytes.

## Testing

If we comment out file removal from fetch flow, the success fetch test case will fail with a leftover file.

### DIFF
--- a/tests/fetch_test.go
+++ b/tests/fetch_test.go
@@ -12,12 +12,12 @@ import (
 	"go.viam.com/test"
 )
 
-// pathForFetchCmd constructs the .mp4 path based on the "from" field in the command
-func pathForFetchCmd(prefix string, cmd map[string]interface{}) string {
-	return "/tmp/" + prefix + "_" + cmd["from"].(string) + ".mp4"
+// pathForFetchCmd constructs the .mp4 path based on the "from" field in the command.
+func pathForFetchCmd(cmd map[string]interface{}) string {
+	return "/tmp/" + videoStoreComponentName + "_" + cmd["from"].(string) + ".mp4"
 }
 
-// assertNoFile checks that the given file path does NOT exist
+// assertNoFile checks that the given file path does NOT exist.
 func assertNoFile(t *testing.T, filePath string) {
 	_, err := os.Stat(filePath)
 	test.That(t, os.IsNotExist(err), test.ShouldBeTrue)
@@ -132,7 +132,7 @@ func TestFetchDoCommand(t *testing.T) {
 		video, ok := res["video"].(string)
 		test.That(t, ok, test.ShouldBeTrue)
 		test.That(t, video, test.ShouldNotBeEmpty)
-		filePath := pathForFetchCmd(videoStoreComponentName, fetchCmd1)
+		filePath := pathForFetchCmd(fetchCmd1)
 		assertNoFile(t, filePath)
 	})
 
@@ -147,7 +147,7 @@ func TestFetchDoCommand(t *testing.T) {
 		_, err = vs.DoCommand(timeoutCtx, fetchCmd2)
 		test.That(t, err, test.ShouldNotBeNil)
 		test.That(t, err.Error(), test.ShouldContainSubstring, "grpc")
-		filePath := pathForFetchCmd(videoStoreComponentName, fetchCmd2)
+		filePath := pathForFetchCmd(fetchCmd2)
 		assertNoFile(t, filePath)
 	})
 
@@ -162,7 +162,7 @@ func TestFetchDoCommand(t *testing.T) {
 		_, err = vs.DoCommand(timeoutCtx, fetchCmd3)
 		test.That(t, err, test.ShouldNotBeNil)
 		test.That(t, err.Error(), test.ShouldContainSubstring, "range")
-		filePath := pathForFetchCmd(videoStoreComponentName, fetchCmd3)
+		filePath := pathForFetchCmd(fetchCmd3)
 		assertNoFile(t, filePath)
 	})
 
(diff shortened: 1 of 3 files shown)


tests/fetch_test.go
```diff
@@ -11,6 +12,17 @@ import (
 	"go.viam.com/test"
 )
 
+// pathForFetchCmd constructs the .mp4 path based on the "from" field in the command.
+func pathForFetchCmd(cmd map[string]interface{}) string {
```
nicksanford, line 16: just have this take a string

nicksanford, line 16:
```suggestion
func pathForFetchCmd(fromTimestamp string) string {
```

videostore/videostore.go
```diff
@@ -296,13 +296,20 @@ func (vs *videostore) Fetch(_ context.Context, r *FetchRequest) (*FetchResponse,
...
 	fetchFilePath := generateOutputFilePath(
 		vs.config.Storage.OutputFileNamePrefix,
 		formatDateTimeToString(r.From),
 		"",
 		tempPath)
 
+	// Always attempt to remove the concat file after the operation.
+	// This handles error cases in Concat where it fails in the middle
+	// of writing.
+	defer func() {
```
nicksanford, line 308: set up this defer after `Concat` succeeds, otherwise if `Concat` fails we won't produce a file & this will log an unnecessary warning.

nicksanford: ah I see, never mind

videostore/videostore.go
```diff
@@ -296,13 +296,20 @@ func (vs *videostore) Fetch(_ context.Context, r *FetchRequest) (*FetchResponse,
...
+		if err := os.Remove(fetchFilePath); err != nil {
```
nicksanford, line 309: check if the file exists with `os.Stat` if it doesn't early return, if it does, do the `os.Remove`

videostore/concat.c
```diff
@@ -131,7 +131,7 @@ int video_store_concat(const char *concat_filepath, const char *output_path) {
 
     if (packet->dts <= prevDts) {
       av_log(NULL, AV_LOG_DEBUG,
-             "video_store_concat skipping non monotonically increaseing dts: "
+             "video_store_concat skipping non monotonically increasing dts: "
```
nicksanford, line 134: ty

videostore/videostore.go
```diff
@@ -296,13 +296,20 @@ func (vs *videostore) Fetch(_ context.Context, r *FetchRequest) (*FetchResponse,
...
+			vs.logger.Debugf("failed to delete temporary file (%s): %v", fetchFilePath, err)
+		}
+	}()
 	if err := vs.concater.Concat(r.From, r.To, fetchFilePath); err != nil {
```
nicksanford, line 313: is removing the concat.txt file a follow up?

seanavery: Seems like txt files are already being deleted properly in the `Concat` method. See here: https://github.com/viam-modules/video-store/blob/d1b64160b6992d01c26b8528ab08c0694ebee09e/videostore/concat.go#L86-L88

seanavery: Looks like txt files are already getting cleaned up properly. https://github.com/viam-modules/video-store/blob/d1b64160b6992d01c26b8528ab08c0694ebee09e/videostore/concat.go#L86-L88

videostore/videostore.go
seanavery: Good call. Added stat lookup before Remove attempt.

videostore/videostore.go
```diff
@@ -296,13 +296,20 @@ func (vs *videostore) Fetch(_ context.Context, r *FetchRequest) (*FetchResponse,
...
 		formatDateTimeToString(r.From),
 		"",
 		tempPath)
 
+	// Always attempt to remove the concat file after the operation.
+	// This handles error cases in Concat where it fails in the middle
+	// of writing.
+	defer func() {
+		if err := os.Remove(fetchFilePath); err != nil {
+			vs.logger.Debugf("failed to delete temporary file (%s): %v", fetchFilePath, err)
```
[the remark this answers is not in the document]
seanavery: Had it as debug so would not be noisy in error case. But with stat check first should be no problem warning, changed.

tests/fetch_test.go
seanavery: Good call. Changed.

tests/fetch_test.go
```diff
@@ -11,6 +12,17 @@ import (
 	"go.viam.com/test"
 )
 
+// pathForFetchCmd constructs the .mp4 path based on the "from" field in the command.
+func pathForFetchCmd(fromTimestamp string) string {
+	return "/tmp/" + videoStoreComponentName + "_" + fromTimestamp + ".mp4"
+}
+
+// assertNoFile checks that the given file path does NOT exist.
```
randhid, line 20:
```suggestion
// assertNoFile checks that the given file path does NOT exist.
// used for testing that the temporary files created by fetch are being properly removed.
```

tests/fetch_test.go
```diff
@@ -120,6 +132,8 @@ func TestFetchDoCommand(t *testing.T) {
 		video, ok := res["video"].(string)
 		test.That(t, ok, test.ShouldBeTrue)
 		test.That(t, video, test.ShouldNotBeEmpty)
+		filePath := pathForFetchCmd(fetchCmd1["from"].(string))
+		assertNoFile(t, filePath)
```
randhid, line 136: Is there a point in these tests prior to this line that you should check that a file has been created?

randhid (approved):
One nit and one question about tests.

tests/fetch_test.go
seanavery: I would like to check for files created and deleted, but I would either need to refactor the concat flow to break the steps up for testing. Or, introduce some[ fs watcher ](https://github.com/fsnotify/fsnotify) to get events. I could create a follow up ticket for this.

randhid: I think this is worth a follow-up ticket. Let's make sure code is testable from the get-go as much as we can!

seanavery: Sounds good. https://viam.atlassian.net/browse/RSDK-10259

PR #61, opened 2025-03-17
[RSDK-8761]   Use Unix UTC timestamp for video-store segment file names

hexbabe:
https://viam.atlassian.net/browse/RSDK-8761

# This PR

- Saves all new segments (segments created with the logic in this diff) with Unix time UTC as its file name
- Maintains backwards compatibility with datetime filename segments (ones created by older versions of video-store)
- Is not a breaking change to `save` or `fetch` requests. Users should not have to change how they format their `DoCommand`s. There should be no difference in the videos and filenames they receive back if this is merged.
- Does not add support to make requests in UTC time or receive back filenames in UTC format

# Manual Testing
## Ran `examples/save/save_client.go`
Ran this test mainly to check that async `save` requests with `to = time.Now` works. Ran it for 10-ish minutes with no failures.

## Tested cross-filename-format `save`
On the av-orin-nano-4, I swapped between the latest video-store ver and a local build off this feature branch to create oscillating strftime segments and unix time segments.
Screenshot 2025-03-20 at 10 28 12 AM av-orin-nano-4

I ran this DoCommand through the App UI to test `save` on the same machine.

```
{
        "command": "save",
        "from": "2025-03-20_10-22-02",
        "to": "2025-03-20_10-25-27",
        "metadata": "metadata",
        "async": false
}
```

And received back a good video. "Good" meaning the timestamps are more or less aligned (accounting for [drift](https://viaminc.slack.com/archives/C06KBQEQDNC/p1742480303492919)) and the video length is reasonable.

https://github.com/user-attachments/assets/de34a7a6-f548-460d-8e26-a7c27d261265

## Ran `examples/fetch/fetch_client.go`
`go run ./examples/fetch/fetch_client.go video-store-1 2025-03-20_10-27-26 2025-03-20_10-27-56` on av-orin-nano-4

https://github.com/user-attachments/assets/711bf52e-e1d9-4b12-b9b7-6544677c78cd

Looks good


tests/save_test.go
```diff
@@ -5,13 +5,24 @@ import (
...
+	"strings"
 	"testing"
 	"time"
 
 	"go.viam.com/rdk/components/camera"
 	"go.viam.com/test"
 )
 
+const (
+	timeFormat            = "2006-01-02_15-04-05"
```
hexbabe, line 18: these existed before. just wanted to make them consts for DRY

videostore/utils.go
```diff
@@ -253,31 +257,31 @@ func sortFilesByDate(files []fileWithDate) {
...
-// Returns a time.Time object and an error if the string is not in the correct format.
+// ParseDateTimeString parses a datetime string in our format "2006-01-02_15-04-05" from localtime,
+// converting it to UTC for internal processing.
 func ParseDateTimeString(datetime string) (time.Time, error) {
-	dateTime, err := time.Parse("2006-01-02_15-04-05", datetime)
+	//nolint:gosmopolitan // datetime format timestamps must be parsed into local time before UTC conversion.
+	t, err := time.ParseInLocation("2006-01-02_15-04-05", datetime, time.Local)
 	if err != nil {
-		return time.Time{}, err
+		return time.Time{}, fmt.Errorf("%w: %w", ErrInvalidDatetimeFormat, err)
```
hexbabe, line 282: btw the NickSanfordGPT in my head says that we might want to assert on `time.ParseError` instead of our custom type, but I tried it out for a bit and got stuck on 

```
    utils_test.go:267: Expected error("parsing time "invalid" as "2006-01-02_15-04-05": cannot parse "invalid" as "2006"") to wrap error("parsing time "" as "": cannot parse "" as """) but it didn't.
```

tests/fetch_test.go
```diff
@@ -14,7 +14,13 @@ import (
 
 // pathForFetchCmd constructs the .mp4 path based on the "from" field in the command.
 func pathForFetchCmd(fromTimestamp string) string {
-	return "/tmp/" + videoStoreComponentName + "_" + fromTimestamp + ".mp4"
+	// Parse the datetime format timestamp
+	t, err := time.Parse("2006-01-02_15-04-05", fromTimestamp)
+	if err != nil {
+		return "/tmp/" + videoStoreComponentName + "_" + fromTimestamp + ".mp4"
+	}
+	// Convert to Unix timestamp
```
nicksanford, line 22: Is it documented that when `time.Parse` fails the returned timestamp is valid?

I don't think this is safe behavior.

Does this actually work?

videostore/utils.go
```diff
@@ -253,31 +254,31 @@ func sortFilesByDate(files []fileWithDate) {
...
-	const minParts = 2
 	baseName := filepath.Base(filePath)
-	parts := strings.Split(baseName, "_")
-	if len(parts) < minParts {
-		return time.Time{}, fmt.Errorf("invalid file name: %s", baseName)
+	nameWithoutExt := strings.TrimSuffix(baseName, filepath.Ext(baseName))
+
+	// Unix timestamp case
+	if timestamp, err := strconv.ParseInt(nameWithoutExt, 10, 64); err == nil {
+		return time.Unix(timestamp, 0).UTC(), nil
```
nicksanford, line 266: are we sure that the unix timestamps have second precision?

videostore/videostore.go
```diff
@@ -295,13 +295,19 @@ func (vs *videostore) Segmenter() *RawSegmenter {
 }
 
 func (vs *videostore) Fetch(_ context.Context, r *FetchRequest) (*FetchResponse, error) {
+	// Convert incoming local times to UTC for consistent timestamp handling
+	// All internal operations and stored timestamps are in UTC
+	r.From = r.From.UTC()
```
nicksanford, line 300: Given the code to take the docommand time range strings and turn them into time.Time lives in the caller of Fetch and Save I recommend making this change in the caller rather than here.

videostore/utils.go
hexbabe: I'm quite confident this will return down to the second always. There are other methods for more precision e.g. https://cs.opensource.google/go/go/+/go1.24.1:src/time/time.go;l=1663

tests/fetch_test.go
hexbabe: >I don't think this is safe behavior.

You're right. This branch is never hit. It should be `err == nil`

videostore/videostore.go
hexbabe: Moved the conversion to a new helper `parseTimeRange`

model/camera/utils.go
```diff
@@ -19,21 +20,34 @@ func getHomeDir() (string, error) {
...
-func ToSaveCommand(command map[string]interface{}) (*videostore.SaveRequest, error) {
+// parseTimeRange parses the from/to timestamps from a command.
+func parseTimeRange(command map[string]interface{}) (from, to time.Time, err error) {
 	fromStr, ok := command["from"].(string)
 	if !ok {
-		return nil, errors.New("from timestamp not found")
+		return time.Time{}, time.Time{}, errors.New("from timestamp not found")
 	}
-	from, err := videostore.ParseDateTimeString(fromStr)
+	from, err = videostore.ParseDateTimeString(fromStr)
```
seanavery, line 29: Just to clarify we are still assuming that the caller of fetch/save are providing from/to date-times in the video-store machine's local timezone?

Do we have an upgrade path to eventually allow users format queries in UTC to avoid the complication of having to know target machines timezone?

hexbabe: I can make a ticket rn

hexbabe: https://viam.atlassian.net/browse/RSDK-10317

videostore/utils.go
```diff
@@ -253,31 +254,31 @@ func sortFilesByDate(files []fileWithDate) {
 }
 
 // extractDateTimeFromFilename extracts the date and time from the filename.
+// Returns UTC time regardless of filename format:
+// - Unix timestamp filenames: directly converts to UTC time
+// - Datetime format filenames (YYYY-MM-DD_HH-mm-ss): parses as local time, converts to UTC
 func extractDateTimeFromFilename(filePath string) (time.Time, error) {
```
nicksanford, line 260: can we change all the function parse strings into time.Time to return time.Time in the timezone the string was in:
If a `"2006-01-02_15-04-05"` format string, it should be in local time, if a UNIX timestamp, should be in UTC time.

The caller can then call a `UTC()` to normalize it.

model/camera/utils.go
```diff
@@ -19,21 +20,34 @@ func getHomeDir() (string, error) {
...
+	}
+	to, err = videostore.ParseDateTimeString(toStr)
+	if err != nil {
+		return time.Time{}, time.Time{}, err
 	}
-	to, err := videostore.ParseDateTimeString(toStr)
+	// Convert incoming local times to UTC for consistent timestamp handling
+	// All internal operations and stored timestamps are in UTC
+	from = from.UTC()
+	to = to.UTC()
```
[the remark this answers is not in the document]
nicksanford: also, I now realize that I was wrong to ask you to remove the call to `.UTC()` within the video-store Fetch and Save methods.  

We should indeed do that. It's ok if it's a no-op if the caller provided time in the UTC timezone.

tests/fetch_test.go
```diff
@@ -14,7 +14,13 @@ import (
 
 // pathForFetchCmd constructs the .mp4 path based on the "from" field in the command.
 func pathForFetchCmd(fromTimestamp string) string {
-	return "/tmp/" + videoStoreComponentName + "_" + fromTimestamp + ".mp4"
+	// Parse the datetime format timestamp
+	t, err := time.Parse("2006-01-02_15-04-05", fromTimestamp)
```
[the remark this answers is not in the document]
hexbabe: done

videostore/utils.go
hexbabe: Fixed

model/camera/utils.go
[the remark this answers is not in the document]
hexbabe: >doesn't ParseDateTimeString already call .UTC() on the returned time?

not anymore after the recent change

>We should indeed do that. It's ok if it's a no-op if the caller provided time in the UTC timezone.

Ok changed back

videostore/videostore.go
hexbabe: Undid the .UTC conversion (made it go back in Save and Fetch), but kept the helper for DRY.

tests/fetch_test.go
hexbabe: Removed complexity after [realizing](https://viaminc.slack.com/archives/C06KBQEQDNC/p1742828601511129?thread_ts=1742825554.647749&cid=C06KBQEQDNC)

hexbabe:
## Manual Test again
Screenshot 2025-03-25 at 10 26 52 AM
```
{
        "command": "save",
        "from": "2025-03-25_10-21-27",
        "to": "2025-03-25_10-21-58",
        "metadata": "metadata",
        "async": false
} // cross-segment save on two strftime-named files with `to` on a boundary
```

https://github.com/user-attachments/assets/49717747-c088-449b-a822-6fa1fd32817c

LGTM

videostore/utils.go
```diff
@@ -229,31 +231,25 @@ func sortFilesByDate(files []fileWithDate) {
...
+	// Datetime format case - keep in local time
+	return ParseDateTimeString(nameWithoutExt)
 }
 
-func formatDateTimeToString(dateTime time.Time) string {
-	return dateTime.Format("2006-01-02_15-04-05")
+// ParseDateTimeString parses a datetime string in our format (2006-01-02_15-04-05)
+// and returns it in local time.
+func ParseDateTimeString(datetime string) (time.Time, error) {
+	//nolint:gosmopolitan // datetime format timestamps must be parsed into local time
```
seanavery, line 251: Do we still need the `nolint` here?

seanavery (approved):

nicksanford (approved):
also tested manually. 

Looks good.

Thanks for getting this in.

videostore/videostore.go
```diff
@@ -43,6 +43,8 @@ const (
 	tempPath              = "/tmp"
 
 	mimeTypeYUYV = "image/yuyv422"
+	// TimeFormat is how we format the timestamp in output filenames and do commands.
+	TimeFormat = "2006-01-02_15-04-05"
```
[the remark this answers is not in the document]
hexbabe: Oh nevermind. It needs to be public bc our tests are a diff pkg

PR #65, opened 2025-04-24
[RSDK-10567]   Add get-storage-status DoCommand to video-store

hexbabe:
[RSDK-10567](https://viam.atlassian.net/browse/RSDK-10567)


videostore/videostore.go
```diff
@@ -497,3 +518,79 @@ func (vs *videostore) Close() {
...
+		start := file.startTime
+		end := start.Add(info.duration)
+
+		if !seenFirstSegment {
+			currentFrom, currentTo = start, end
+			seenFirstSegment = true
+			continue
+		}
+
+		if !start.After(currentTo) {
```
seanavery, line 565: Is this essentially a 1 second buffer? Or this expected to hit on every new segment?

seanavery (commented):
Looking great!

Couple of follow ups:
- README documentation for `GetStorageState`
- Wondering if any of the logic here can be DRY'd with the [cleanupStorage](https://github.com/viam-modules/video-store/blob/main/videostore/videostore.go#L431) helper in video-store.

videostore/videostore.go
```diff
@@ -497,3 +518,79 @@ func (vs *videostore) Close() {
...
+	}
+
+	var (
+		ranges                 []VideoTimeRange
+		currentFrom, currentTo time.Time
+		seenFirstSegment       bool
+	)
+
+	for _, file := range files {
+		info, err := getVideoInfo(file.name)
```
nicksanford, line 550:
```suggestion
		videoInfo, err := getVideoInfo(file.name)
```

videostore/videostore.go
```diff
@@ -497,3 +518,79 @@ func (vs *videostore) Close() {
...
+			From: currentFrom.UTC().Format(TimeFormat) + "Z",
+			To:   currentTo.UTC().Format(TimeFormat) + "Z",
+		})
+	}
+
+	return &StorageState{
+		DiskUsage: DiskUsageState{
+			StorageUsedGB:            usedGB,
+			StorageLimitGB:           limitGB,
+			DeviceStorageRemainingGB: remainingGB,
```
nicksanford, line 592: are you sure that this will return the amount of storage space on the disk partition the ~/.viam/video-store directory has available to it and not the sum total of storage space ALL partitions have available to them?

videostore/videostore.go
```diff
@@ -497,3 +518,79 @@ func (vs *videostore) Close() {
...
+	var (
+		ranges                 []VideoTimeRange
+		currentFrom, currentTo time.Time
+		seenFirstSegment       bool
+	)
+
+	for _, file := range files {
+		info, err := getVideoInfo(file.name)
+		if err != nil {
+			vs.logger.Warnf("skipping unfinished or corrupt file %s: %v", file.name, err)
```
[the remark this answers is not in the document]
hexbabe: Yep that makes sense

videostore/videostore.go
nicksanford: I expect this to trigger a bunch of false positives given if there is even a few milliseconds missing between one segement file and another we are going to create a new `VideoTimeRange`

I recommend providing a little bit of slop (lets say 1 second) that can be missing from one file starting and another file ending and we can still consider the video segments contiguous.

tests/storage_state_test.go
```diff
@@ -0,0 +1,137 @@
...
+	test.That(t, ok, test.ShouldBeTrue)
+	test.That(t, limit, test.ShouldEqual, float64(10))
+	remaining, ok := disk["device_storage_remaining_gb"].(float64)
+	test.That(t, ok, test.ShouldBeTrue)
+	test.That(t, remaining, test.ShouldBeGreaterThan, 0.0)
+
+	// Validate stored_video ranges
+	videoList, ok := res["stored_video"].([]interface{})
+	test.That(t, ok, test.ShouldBeTrue)
+	test.That(t, len(videoList), test.ShouldBeGreaterThan, 0)
```
nicksanford, line 117: is there a specific number of `videoList` elements we should expect?

hexbabe: Yes. 7 contiguous segments in our artifacts plus 1 more we wait on created in realtime during the unit test run. So 8 should be the number. I can change.

model/camera/camera.go
```diff
@@ -152,6 +154,45 @@ func (c *component) DoCommand(ctx context.Context, command map[string]interface{
...
+		storagePath := vsConfig.Storage.StoragePath
+		storageLimitGB := vsConfig.Storage.SizeGB
+
+		fsUsage, err := diskusage.Statfs(storagePath)
+		var remainingGB float64
+		if err != nil {
+			c.logger.Warnw("failed to get filesystem stats for remaining space", "path", storagePath, "error", err)
+		} else {
+			remainingGB = float64(fsUsage.AvailableBytes) / float64(gigabyte)
+		}
```
nicksanford, line 174: Have you observed this failing?  I think we should return an error if this happens. Otherwise the caller won't be able to tell the difference between no remaining disk space and remaining disk space, but a failed Statfs call...

model/camera/camera.go
```diff
@@ -152,6 +154,45 @@ func (c *component) DoCommand(ctx context.Context, command map[string]interface{
...
+		state, err := c.videostore.GetStorageState()
+		if err != nil {
+			return nil, err
+		}
+
+		vsConfig := c.videostore.GetConfig()
+		storagePath := vsConfig.Storage.StoragePath
+		storageLimitGB := vsConfig.Storage.SizeGB
+
+		fsUsage, err := diskusage.Statfs(storagePath)
```
nicksanford, line 168: can we move this into `videostore.GetStorageState`?

I'd like DoCommand to be:
```
switch command {
case "command-string":
		state, err := c.videostore.GetStorageState()
		if err != nil {
			return nil, err
		}
                 return formatState(state), nil

...
}
```

videostore/indexer.go
```diff
@@ -0,0 +1,406 @@
...
+
+	if len(filePaths) == 0 {
+		return nil
+	}
+
+	tx, err := ix.db.Begin()
+	if err != nil {
+		return fmt.Errorf("failed to begin transaction: %w", err)
+	}
+	defer tx.Rollback() // rollback on failure
```
nicksanford, line 227:
```suggestion
         g := utils.NewGuard(tx.Rollback)
	...
	g.Success()
	return tx.Commit()
```

videostore/indexer.go
```diff
@@ -0,0 +1,406 @@
...
+
+	query := fmt.Sprintf("DELETE FROM %s WHERE file_path = ?", segmentsTableName)
```
nicksanford, line 229:
```suggestion
	query := fmt.Sprintf("DELETE FROM %s WHERE file_path = ?;", segmentsTableName)
```

videostore/indexer.go
```diff
@@ -0,0 +1,406 @@
...
+	ix.mu.Lock()
+	defer ix.mu.Unlock()
+
+	var state StorageState
+
+	// Query all segments, ordered by start time
+	query := fmt.Sprintf(`
+	SELECT start_time_unix, duration_ms, size_bytes
+	FROM %s
+	ORDER BY start_time_unix ASC
```
nicksanford, line 284: this is going to use string sort... does this actually work?

videostore/indexer.go
```diff
@@ -0,0 +1,406 @@
...
+		if err := rows.Scan(&startTimeUnix, &durationMs, &sizeBytes); err != nil {
+			return StorageState{}, fmt.Errorf("failed to scan segment row: %w", err)
+		}
+
+		state.SegmentCount++
+		if durationMs.Valid {
+			state.TotalDurationMs += durationMs.Int64
+		} else {
+			ix.logger.Warnw("segment found with NULL duration in index", "start_time_unix", startTimeUnix)
+			continue
```
nicksanford, line 311: have you observed this condition? When does it happen?

videostore/indexer.go
```diff
@@ -0,0 +1,406 @@
...
+		}
+
+		if sizeBytes.Valid {
+			segments = append(segments, SegmentMetadata{
+				FilePath:  filePath,
+				SizeBytes: sizeBytes.Int64,
+			})
+		} else {
+			ix.logger.Warnw("segment found with NULL size in index, skipping", "file_path", filePath)
```
nicksanford, line 379: when does this happen?

videostore/indexer.go
```diff
@@ -0,0 +1,406 @@
...
+
+	var segments []SegmentMetadata
+
+	for rows.Next() {
+		var filePath string
+		var sizeBytes sql.NullInt64
+
+		if err := rows.Scan(&filePath, &sizeBytes); err != nil {
+			ix.logger.Errorw("failed to scan segment row during full query", "error", err)
+			continue
```
nicksanford, line 370: why not return this as an error?

hexbabe: It is so one single problematic row doesn't affect us returning *something* to the user.

videostore/indexer.go
hexbabe: I haven't seen it happen. I thought it would be best practice to case it out.

videostore/indexer.go
hexbabe: Same thing as https://github.com/viam-modules/video-store/pull/65#discussion_r2086864562

>I haven't seen it happen. I thought it would be best practice to case it out.

videostore/indexer.go
```diff
@@ -0,0 +1,406 @@
...
+		}
+		ix.logger.Debugf("removed indexed segment: %s", path)
+	}
+
+	return tx.Commit()
+}
+
+// getVideoDuration uses the getVideoInfo utility to get the duration of a video file.
+// This function does not require the indexer mutex.
+func (ix *Indexer) getVideoDuration(filePath string) (int64, error) {
```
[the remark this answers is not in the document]
hexbabe: Done

videostore/indexer.go
```diff
@@ -0,0 +1,406 @@
...
+	for path, info := range diskFiles {
+		if _, exists := indexedFiles[path]; !exists {
+			ix.logger.Debugf("found new segment to index: %s", path)
+			if err := ix.indexNewFile(path, info); err != nil {
+				ix.logger.Errorw("failed to index new file", "path", path, "error", err)
+			}
+		}
+	}
+
+	return nil
```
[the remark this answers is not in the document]
hexbabe: From our meeting and that diagram, I assumed the deleter handles that exclusively i.e. polling shouldn't handle removals from index or filesys

videostore/indexer.go
```diff
@@ -0,0 +1,406 @@
...
+}
+
+// getIndexedFiles retrieves the set of file paths currently in the index.
+// Assumes ix.mu is held (or called from a context where the lock is held).
+func (ix *Indexer) getIndexedFiles() (map[string]struct{}, error) {
+	query := "SELECT file_path FROM " + segmentsTableName
```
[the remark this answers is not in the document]
hexbabe: Done

videostore/indexer.go
```diff
@@ -0,0 +1,406 @@
...
+
+	query := fmt.Sprintf(`
+	CREATE TABLE IF NOT EXISTS %s (
+		file_path TEXT PRIMARY KEY,
+		start_time_unix INTEGER NOT NULL,
+		duration_ms INTEGER,
+		size_bytes INTEGER,
+		creation_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+	);
+	CREATE INDEX IF NOT EXISTS idx_start_time ON %s (start_time_unix);
```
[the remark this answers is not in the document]
hexbabe: This is what I want to figure out today (+ ensuring correctness)

model/camera/camera.go
hexbabe: I have not. I agree this is severe enough to make it an error.

videostore/indexer.go
```diff
@@ -0,0 +1,388 @@
...
+			}
+			fullPath := filepath.Join(ix.storagePath, entry.Name())
+			files[fullPath] = info
+		}
+	}
+	return files, nil
+}
+
+// indexNewFile extracts metadata and adds a new file to the index.
+func (ix *Indexer) indexNewFile(filePath string, fileInfo os.FileInfo) error {
```
nicksanford, line 179: you don't need the full `os.FileInfo` just take the size.

videostore/indexer.go
```diff
@@ -0,0 +1,388 @@
...
+	ix.logger.Debugw("indexed new file", "file", filePath, "start_time", startTimeStr, "duration_ms", durationMs, "size_bytes", sizeBytes)
+	return nil
+}
+
+// RemoveIndexedFiles removes file paths from the index.
+func (ix *Indexer) RemoveIndexedFiles(filePaths []string) error {
+	if len(filePaths) == 0 {
+		return nil
+	}
+
```
nicksanford, line 210: why do you need to delete all the filePaths in a single transaction?

videostore/indexer.go
```diff
@@ -0,0 +1,388 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to query all segments: %w", err)
+	}
+	defer rows.Close()
+
+	var segments []SegmentMetadata
+
+	for rows.Next() {
+		var filePath string
+		var sizeBytes sql.NullInt64
```
nicksanford, line 352: should not be nullable

videostore/indexer.go
nicksanford: but then we are lying and returning incomplete data.

If we don't know specifically why this error would happen for one row and not all rows, lets log which row we failed on and fail the entire request if we hit an error.

videostore/indexer.go
```diff
@@ -0,0 +1,388 @@
...
+
+		if err := rows.Scan(&filePath, &sizeBytes); err != nil {
+			ix.logger.Errorw("failed to scan segment row during full query", "error", err)
+			continue
+		}
+
+		if sizeBytes.Valid {
```
nicksanford, line 359: this should always be valid, don't insert invalid data into the db

videostore/indexer.go
```diff
@@ -0,0 +1,388 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, entry := range entries {
+		if !entry.IsDir() && strings.HasSuffix(entry.Name(), ".mp4") {
+			info, err := entry.Info()
+			if err != nil {
+				ix.logger.Warnw("failed to get FileInfo for disk file", "name", entry.Name(), "error", err)
+				continue
```
nicksanford, line 169: lets not continue here, lets return an error. Unless we know of a specific reason why this would happen we should bias towards not omitting information on error.

videostore/videostore.go
```diff
@@ -65,6 +65,7 @@ type VideoStore interface {
 	Save(ctx context.Context, r *SaveRequest) (*SaveResponse, error)
 	Close()
 	GetStorageState() (*StorageState, error)
+	GetConfig() Config
```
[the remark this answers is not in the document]
hexbabe: Good call I will consolidate

model/camera/camera.go
```diff
@@ -158,16 +160,32 @@ func (c *component) DoCommand(ctx context.Context, command map[string]interface{
...
+		videoList := make([]map[string]interface{}, 0, len(state.Ranges))
+		for _, timeRange := range state.Ranges {
+			fromStr := timeRange.Start.UTC().Format(videostore.TimeFormat) + "Z"
+			toStr := timeRange.End.UTC().Format(videostore.TimeFormat) + "Z"
 			videoList = append(videoList, map[string]interface{}{
-				"from": r.From,
-				"to":   r.To,
+				"from": fromStr,
+				"to":   toStr,
 			})
```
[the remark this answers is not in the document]
hexbabe: Got it thanks

videostore/indexer.go
```diff
@@ -0,0 +1,406 @@
+package videostore
```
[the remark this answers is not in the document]
hexbabe: Gotcha. Privated all.

model/camera/camera.go
hexbabe: Changed to return error on statfs failure

model/camera/camera.go
hexbabe: Yes it has been moved

videostore/indexer.go
```diff
@@ -0,0 +1,406 @@
...
+	if err := ix.initializeDB(); err != nil {
+		err := ix.Close() // close out db if initialization fails
+		if err != nil {
+			logger.Errorw("error closing indexer", "error", err)
+		}
+		return nil, fmt.Errorf("failed to initialize index db schema: %w", err)
+	}
+
+	return ix, nil
+}
```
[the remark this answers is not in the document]
hexbabe: I added `setup` and call it after construction for all calls

videostore/indexer.go
```diff
@@ -0,0 +1,388 @@
...
+// It interacts with the filesystem only and does not require the indexer mutex.
+func (ix *Indexer) getDiskFiles() (map[string]os.FileInfo, error) {
+	files := make(map[string]os.FileInfo)
+	entries, err := os.ReadDir(ix.storagePath)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, entry := range entries {
+		if !entry.IsDir() && strings.HasSuffix(entry.Name(), ".mp4") {
```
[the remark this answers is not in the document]
hexbabe: Added a const and used that instead

videostore/indexer.go
hexbabe: Changed to err return

videostore/videostore.go
```diff
@@ -486,14 +555,56 @@ func (vs *videostore) asyncSave(ctx context.Context, from, to time.Time, path st
...
+	if len(errs) > 0 {
+		errMsg := "errors during close: "
+		for i, err := range errs {
+			if i > 0 {
+				errMsg += ", "
+			}
+			errMsg += err.Error()
+		}
+		vs.logger.Errorw(errMsg)
+	}
```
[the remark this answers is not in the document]
hexbabe: Deleted

videostore/videostore.go
```diff
@@ -419,45 +454,79 @@ func (vs *videostore) deleter(ctx context.Context) {
...
+		vs.logger.Warn("cleanup required based on disk size, but indexer found no segments. Index might be corrupted or empty.")
+		return nil
+	}
+
+	vs.logger.Debugf("retrieved %d segments from indexer, starting deletion process", len(allSegments))
+
+	// Iterate over the segments, deleting oldest first until enough space is freed
+	for _, segment := range allSegments {
+		vs.logger.Debugf("Attempting to delete oldest segment: %s (size: %d bytes)", segment.FilePath, segment.SizeBytes)
+		err := os.Remove(segment.FilePath)
```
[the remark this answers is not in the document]
nicksanford: Good point. Move it into the indexer background goroutine

videostore/indexer.go
```diff
@@ -0,0 +1,392 @@
...
+	ix.setupDone = true
+	return nil
+}
+
+func (ix *indexer) initializeDB() error {
+	query := fmt.Sprintf(`
+	CREATE TABLE IF NOT EXISTS %s (
+		file_path TEXT PRIMARY KEY,
+		start_time_unix INTEGER NOT NULL,
+		duration_ms INTEGER,
```
seanavery, line 79: Could we consider preemptively adding rows for the rest of `videoInfo`(width, height, and codec)

```type videoInfo struct {
	duration time.Duration
	width    int
	height   int
	codec    string
}
```

The utility [matchStorageToRange](https://github.com/viam-modules/video-store/blob/main/videostore/utils.go#L269) could be refactored a bit to use the indexer instead.

I think this could be a follow-up ticket, but not sure about the complications of migrating schemas once things go live.

hexbabe: I'm down if it makes things easier down the line

videostore/indexer.go
nicksanford: No. I think you can just delete them individually. I don't think there is a reason why you need to atomically delete all the files together.

Do you know of one?

hexbabe: Attempting atomicity was the main one. In case someone unplugs the machine etc.

videostore/indexer.go
```diff
@@ -0,0 +1,406 @@
...
+	defer ix.mu.Unlock()
+
+	query := fmt.Sprintf(`
+	CREATE TABLE IF NOT EXISTS %s (
+		file_path TEXT PRIMARY KEY,
+		start_time_unix INTEGER NOT NULL,
+		duration_ms INTEGER,
+		size_bytes INTEGER,
+		creation_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+	);
```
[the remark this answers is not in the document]
hexbabe: is this in regards to it being the primary key?

videostore/indexer.go
hexbabe: Changed to error return

videostore/indexer.go
```diff
@@ -0,0 +1,406 @@
...
+				ix.logger.Errorw("error during indexer scan", "error", err)
+			}
+		}
+	}
+}
+
+// scanAndIndex performs one cycle of scanning the storage directory and updating the index.
+// It focuses on adding new files found on disk to the index.
+// It does not remove index entries for deleted disk files; that is handled by the cleanup/deletion logic.
+func (ix *Indexer) scanAndIndex() error {
```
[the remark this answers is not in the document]
hexbabe: The former `getDiskFiles, getIndexedFiles` do not take long to run 
```
5/14/2025, 3:37:18 PM debug rdk:component:camera/video_camera-EL5c     videostore/indexer.go:151 getIndexedFiles took 372.781µs   log_ts 2025-05-14T19:37:18.007Z

5/14/2025, 3:37:18 PM debug rdk:component:camera/video_camera-EL5c     videostore/indexer.go:144 getDiskFiles took 51.95095ms   log_ts 2025-05-14T19:37:18.005Z
```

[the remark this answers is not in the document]
hexbabe: I also observed by calling `get-storage-state` repeatedly that as the indexer does its first monster run, the storage_used_gb field slowly grows in size:
```
  "disk_usage": {
    "device_storage_remaining_gb": 835.1371192932129,
    "storage_path": "/root/.viam/video-storage/video_camera-EL5c",
    "storage_used_gb": 37.00883391499519, // growing per call
    "storage_limit_gb": 100
  },
```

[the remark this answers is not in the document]
hexbabe: `5/14/2025, 3:55:31 PM debug rdk:component:camera/video_camera-EL5c     videostore/indexer.go:188 indexing took 18m13.250427453s   log_ts 2025-05-14T19:55:31.258Z`

18 minutes is a bit crazy

videostore/indexer.go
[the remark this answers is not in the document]
hexbabe: sorry I haven't check this yet. I will tomorrow

videostore/indexer.go
```diff
@@ -0,0 +1,451 @@
...
+		}
+		if _, exists := indexedFiles[path]; exists {
+			continue
+		}
+		if !oldestDBTime.IsZero() {
+			fileTime, err := extractDateTimeFromFilename(path)
+			if err != nil {
+				return fmt.Errorf("failed to extract timestamp from filename, skipping: %w", err)
+			}
+			if fileTime.Before(oldestDBTime) {
```
nicksanford, line 174: so... consider the following case:
1. the module starts indexing
2. it get half way through the files on disk
3. it crashes
4. it reboots
5. It never indexes the files that were on disk when it booted which it didn't index yet...

videostore/indexer.go
```diff
@@ -0,0 +1,451 @@
...
+	}
+	durationMs := info.duration.Milliseconds()
+
+	query := fmt.Sprintf("INSERT INTO %s (file_path, start_time_unix, duration_ms, size_bytes) VALUES (?, ?, ?, ?);", segmentsTableName)
+	_, err = ix.db.Exec(query, filePath, startTime.Unix(), durationMs, fileSize)
+	if err != nil {
+		return fmt.Errorf("failed to insert segment into index: %w", err)
+	}
+
+	startTimeStr := startTime.UTC().Format(TimeFormat) + "Z"
```
nicksanford, line 206: this should be done with a function

videostore/indexer.go
```diff
@@ -0,0 +1,451 @@
...
+	args := make([]interface{}, len(paths))
+	for i, path := range paths {
+		placeholders[i] = "?"
+		args[i] = path
+	}
+	//nolint:gosec // this is safe because segmentsTableName is a constant and placeholders are '?'
+	query := fmt.Sprintf("DELETE FROM %s WHERE file_path IN (%s) RETURNING file_path", segmentsTableName, strings.Join(placeholders, ", "))
+
+	// QueryContext is used because the DELETE statement includes a RETURNING clause to get the deleted file paths.
+	rows, err := ix.db.QueryContext(ctx, query, args...)
```
nicksanford, line 277: why can't you pass in `paths...`?

videostore/indexer.go
```diff
@@ -0,0 +1,451 @@
...
+
+	// Build the SQL query
+	placeholders := make([]string, len(paths))
+	args := make([]interface{}, len(paths))
+	for i, path := range paths {
+		placeholders[i] = "?"
+		args[i] = path
+	}
+	//nolint:gosec // this is safe because segmentsTableName is a constant and placeholders are '?'
+	query := fmt.Sprintf("DELETE FROM %s WHERE file_path IN (%s) RETURNING file_path", segmentsTableName, strings.Join(placeholders, ", "))
```
nicksanford, line 274:
```suggestion
	query := fmt.Sprintf("DELETE FROM %s WHERE file_path IN (%s) RETURNING file_path;", segmentsTableName, strings.Join(placeholders, ", "))
```

videostore/indexer.go
hexbabe: `cannot use paths (variable of type []string) as []any value in argument to ix.db.QueryContext`

videostore/indexer.go
hexbabe: Added a helper to utils.go

videostore/indexer.go
```diff
@@ -0,0 +1,451 @@
...
+		return nil, nil
+	}
+	bytesToDelete := currentSizeBytes - maxStorageSizeBytes
+
+	allSegments, err := ix.getSegmentsAscTime(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if len(allSegments) == 0 {
+		return nil, errors.New("cleanup required based on disk size, but indexer found no segments")
```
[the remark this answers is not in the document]
hexbabe: Good point. I think I didn't hit this during manual testing bc it's non-deterministic e.g. we hover between being over the limit and not quite at the limit for these 100gb cameras based on the deletion freq and timing.

videostore/indexer.go
```diff
@@ -0,0 +1,451 @@
...
+	}
+	indexedFiles, err := ix.getIndexedFiles(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Query for the oldest DB file in order to determine if a file is new
+	var oldestDBFile string
+	row := ix.db.QueryRowContext(ctx, fmt.Sprintf("SELECT file_path FROM %s ORDER BY start_time_unix ASC LIMIT 1", segmentsTableName))
+	if err := row.Scan(&oldestDBFile); err != nil && err != sql.ErrNoRows {
```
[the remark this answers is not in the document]
hexbabe: in that case, we have no records in the db and both `oldestDBFile` and `oldestDBTime` are their default zero falsey values. For time, this is epoch, so nothing will be older. This is handled appropriately though I thought... do you see an issue?

videostore/indexer.go
hexbabe: Synced offline and arrived at the current solution of marking as deleted in the db then actually deleting later

videostore/indexer.go
```diff
@@ -0,0 +1,451 @@
...
+			ix.logger.Infof("deleted file from disk: %s", path)
+		} else {
+			ix.logger.Warnw("failed to delete file from disk", "file", path, "error", err)
+		}
+	}
+
+	// 3. Index new files
+	if err := ix.indexNewFiles(ctx); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
hexbabe: Got it. Swapped order and changed signature

tests/storage_state_test.go
```diff
@@ -0,0 +1,156 @@
...
+		test.That(t, ok, test.ShouldBeTrue)
+
+		test.That(t, strings.HasSuffix(fromStr, "Z"), test.ShouldBeTrue)
+		test.That(t, strings.HasSuffix(toStr, "Z"), test.ShouldBeTrue)
+
+		fromTime, err := time.Parse(videostore.TimeFormat, strings.TrimSuffix(fromStr, "Z"))
+		test.That(t, err, test.ShouldBeNil)
+		toTime, err := time.Parse(videostore.TimeFormat, strings.TrimSuffix(toStr, "Z"))
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, fromTime.Before(toTime), test.ShouldBeTrue)
```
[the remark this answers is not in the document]
hexbabe: yes added, except for `between the start time of the test and the current time` because these segments are pulled from google cloud and from 2024.

model/camera/camera.go
```diff
@@ -152,6 +153,35 @@ func (c *component) DoCommand(ctx context.Context, command map[string]interface{
...
+		disk := map[string]interface{}{
+			"storage_used_gb":             float64(state.TotalSizeBytes) / float64(gigabyte),
+			"storage_limit_gb":            float64(state.StorageLimitGB),
+			"device_storage_remaining_gb": state.DeviceStorageRemainingGB,
+			"storage_path":                state.StoragePath,
+		}
+
+		videoList := make([]map[string]interface{}, 0, len(state.Ranges))
+		for _, timeRange := range state.Ranges {
+			fromStr := timeRange.Start.UTC().Format(videostore.TimeFormat) + "Z"
```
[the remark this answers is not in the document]
hexbabe: yes made one

tests/storage_state_test.go
```diff
@@ -0,0 +1,157 @@
...
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/viam-modules/video-store/videostore"
+	"go.viam.com/rdk/components/camera"
+	"go.viam.com/test"
+)
+
+func TestGetStorageStateDoCommand(t *testing.T) {
```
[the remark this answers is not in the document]
hexbabe: this has been added

videostore/indexer.go
```diff
@@ -0,0 +1,512 @@
...
+		return nil, fmt.Errorf("error iterating over all segments query result: %w", err)
+	}
+
+	ix.logger.Debugf("retrieved %d segments from index", len(segments))
+	return segments, nil
+}
+
+func (ix *indexer) close() error {
+	if ix.setupDone {
+		err := ix.db.Close()
```
nicksanford, line 504: What happens if db has a method called on it after `Close()` is called?

If the answer is:
1. You get an error, but no crash / no undefined behavior, and assuming all code that calls db methods handles errors (without crashing etc) then making setupDone atomic should be sufficient
2. If it's undefined behavior / crashes / hangs / another unacceptable condition. We should wrap setupDone & db in a rw mutex

If 2:
1. Make a closedCtx on the struct
2. At beginning of close cancel that context
3. make a rw mutex on the struct that protects the db, write in close read everywhere elase
4. check closedCtx within the rw mutex / pass the decendent context to all the sqlite functions to see if close is trying to terminate the indexer

hexbabe:
```
2025-05-16T17:10:31.731Z	INFO	rdk:component:camera/video-store-1	videostore/indexer.go:526	DB: got db error after close: sql: database is closed
```

```
func (ix *indexer) close() error {
	if ix.setupDone.Load() {
		err := ix.db.Close()
		_, errDB := ix.db.Query("SELECT 1")
		if errDB != nil {
			ix.logger.Infof("DB: got db error after close: %v", errDB)
		} else {
			ix.logger.Error("DB: db connection still open after close")
		}
		if err != nil {
			return fmt.Errorf("error closing index db: %w", err)
		}
		ix.setupDone.Store(false)
		ix.logger.Debug("indexer closed")
	}
	return nil
}
```

looks like 1 is true

videostore/indexer.go
```diff
@@ -0,0 +1,451 @@
...
+				videoRanges.Ranges = append(videoRanges.Ranges, *currentRange)
+				currentRange = &videoRange{Start: segmentStart, End: segmentEnd}
+			} else {
+				currentRange.End = segmentEnd
+			}
+		}
+	}
+	if currentRange != nil {
+		videoRanges.Ranges = append(videoRanges.Ranges, *currentRange)
+	}
```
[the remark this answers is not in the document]
hexbabe: yes done

videostore/indexer.go
```diff
@@ -0,0 +1,451 @@
...
+		return fmt.Errorf("failed to initialize index db schema: %w", err)
+	}
+	ix.setupDone = true
+	return nil
+}
+
+func (ix *indexer) initializeDB() error {
+	query := fmt.Sprintf(`
+	CREATE TABLE IF NOT EXISTS %s (
+		file_path TEXT PRIMARY KEY,
```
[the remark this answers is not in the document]
hexbabe: Gotcha changed

model/camera/camera.go
```diff
@@ -152,6 +153,35 @@ func (c *component) DoCommand(ctx context.Context, command map[string]interface{
...
+			videoList = append(videoList, map[string]interface{}{
+				"from": fromStr,
+				"to":   toStr,
+			})
+		}
+
+		return map[string]interface{}{
+			"command":      "get-storage-state",
+			"disk_usage":   disk,
+			"stored_video": videoList,
```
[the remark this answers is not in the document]
hexbabe: yes added a func for formatting the resp

videostore/indexer.go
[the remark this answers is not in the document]
hexbabe: Synced offline and the performance difference is better without the index. Removing.

videostore/encoder.go
```diff
@@ -78,7 +79,7 @@ func (e *encoder) initialize() error {
 
 	if ret != C.VIDEO_STORE_ENCODER_RESP_OK {
 		err := errors.New("failed to initialize encoder")
-		e.logger.Errorf("%s: %d: %s", err.Error(), ret, ffmpegError(ret))
+		e.logger.Errorf("%s: %d: %s", err.Error(), ret, vsutils.FFmpegError(int(ret)))
```
hexbabe, line 82: had to change the param of `ffmpegError` which is now `vsutils.FFmpegError` to take in a Go int due to `cannot use C.int(ret) (value of type _Ctype_int) as vsutils._Ctype_int value in argument to vsutils.FFmpegError`

videostore/videostore.go
```diff
@@ -25,13 +31,9 @@ const (
 	segmentSeconds = 30 // seconds
 	videoFormat    = "mp4"
 
-	deleterInterval = 1  // minutes
-	retryInterval   = 1  // seconds
-	asyncTimeout    = 60 // seconds
-	tempPath        = "/tmp"
-
-	// TimeFormat is how we format the timestamp in output filenames and do commands.
```
hexbabe, line 33: moved to `vsutils`

Makefile
```diff
@@ -57,18 +57,18 @@ endif
 ifeq ($(SOURCE_OS),darwin)
 	SUBST = $(HOMEBREW_PREFIX)/Cellar/x264/r3108/lib/libx264.a
 endif
-CGO_LDFLAGS = $(subst -lx264, $(SUBST),$(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) pkg-config --libs $(FFMPEG_LIBS))) 
+CGO_LDFLAGS = $(subst -lx264, $(SUBST),$(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) pkg-config --libs $(FFMPEG_LIBS))) -ldl
```
hexbabe, line 60: adding `-ldl` since our C files our no longer in a flat directory structure

videostore/indexer.go
seanavery: @nicksanford What do you think about adding additional videoInfo attrs to the database?

videostore/indexer/indexer.go
```diff
@@ -0,0 +1,540 @@
...
+	vsutils "github.com/viam-modules/video-store/videostore/utils"
+	"go.viam.com/rdk/logging"
+)
+
+const (
+	dbFileName             = "index.sqlite.db"
+	dbFileMode             = 0o750
+	segmentsTableName      = "segments"
+	videoFileSuffix        = ".mp4"
+	indexerRefreshInterval = 10 * time.Second
```
seanavery, line 28: @nicksanford @hexbabe What do we think about bumping the polling time down to 1 second?

At the current 10 second interval, we will have a 1/3 chance of inaccurately reporting the most recent segment.

videostore/indexer/indexer_test.go
```diff
@@ -0,0 +1,470 @@
+package indexer
```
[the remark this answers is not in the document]
nicksanford: eh, not important

videostore/indexer/indexer.go
```diff
@@ -0,0 +1,540 @@
...
+func (ix *Indexer) markSegmentsAsDeleted(ctx context.Context, names []string) error {
+	if len(names) == 0 {
+		return nil
+	}
+
+	// Build up query
+	currentTime := time.Now().UTC()
+	placeholders := make([]string, len(names))
+	args := make([]interface{}, len(names)+1) // +1 for currentTime
+	args[0] = currentTime
```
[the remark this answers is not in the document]
hexbabe: `currentTime` is needed because it's the value being assigned to the `deleted_at` for the matched rows.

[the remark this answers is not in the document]
hexbabe: I don't think it affects affect rows due to it not being an item in the WHERE clause. Let me know if my understanding of this SQL query is incorrect...

videostore/indexer/indexer_test.go
```diff
@@ -0,0 +1,470 @@
...
+	err := os.MkdirAll(storagePath, 0o750)
+	test.That(t, err, test.ShouldBeNil)
+
+	// Use a unique in-memory database name for each test to prevent interference.
+	dbName := fmt.Sprintf("file:%s_db?mode=memory&cache=shared", strings.ReplaceAll(t.Name(), "/", "_"))
+	db, err := sql.Open("sqlite3", dbName)
+	test.That(t, err, test.ShouldBeNil)
+
+	idx.db = db
+	idx.dbPath = dbName
```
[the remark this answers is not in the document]
hexbabe: done

videostore/indexer/indexer_test.go
```diff
@@ -0,0 +1,470 @@
...
+	err = idx.initializeDB()
+	test.That(t, err, test.ShouldBeNil)
+	idx.setupDone.Store(true)
+
+	cleanup := func() {
+		err := idx.Close()
+		test.That(t, err, test.ShouldBeNil)
+	}
+	return idx, cleanup
+}
```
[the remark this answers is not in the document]
hexbabe: done

videostore/indexer/indexer.go
```diff
@@ -0,0 +1,540 @@
...
+		args[i+1] = name
+	}
+	//nolint:gosec // segmentsTableName is a constant, placeholders are '?'
+	query := fmt.Sprintf(
+		"UPDATE %s SET deleted_at = ? WHERE file_name IN (%s) AND deleted_at IS NULL;",
+		segmentsTableName,
+		strings.Join(placeholders, ", "),
+	)
+
+	result, err := ix.db.ExecContext(ctx, query, args...)
```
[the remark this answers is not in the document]
hexbabe: done

videostore/indexer/indexer.go
```diff
@@ -0,0 +1,540 @@
...
+	indexStart := time.Now()
+	if err := ix.indexNewFiles(ctx); err != nil {
+		ix.logger.Errorw("error during indexNewFiles phase", "error", err)
+	}
+	ix.logger.Debugf("TIMING: indexNewFiles took %v", time.Since(indexStart))
+
+	// 2. Mark files to delete based on storage limits
+	cleanupDBStart := time.Now()
+	if err := ix.cleanupDB(ctx); err != nil {
+		ix.logger.Errorw("error in cleanupDB phase (marking for deletion)", "error", err)
```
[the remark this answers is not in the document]
hexbabe: I think it's reasonable to return after that log. Added

videostore/indexer/indexer.go
```diff
@@ -0,0 +1,540 @@
...
+		return fmt.Errorf("failed to initialize index db schema: %w", err)
+	}
+	ix.setupDone.Store(true)
+	return nil
+}
+
+func (ix *Indexer) initializeDB() error {
+	query := fmt.Sprintf(`
+	CREATE TABLE IF NOT EXISTS %s (
+		id INTEGER PRIMARY KEY AUTOINCREMENT,
```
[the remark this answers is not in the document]
hexbabe: Added

videostore/videostore.go
```diff
@@ -491,9 +471,40 @@ func (vs *videostore) Close() {
 	if vs.workers != nil {
 		vs.workers.Stop()
 	}
+
+	if vs.indexer != nil {
+		if err := vs.indexer.Close(); err != nil {
```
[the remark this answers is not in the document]
hexbabe: yes

videostore/utils/utils.go
```diff
@@ -255,7 +252,12 @@ func ParseDateTimeString(datetime string) (time.Time, error) {
 	return time.ParseInLocation(TimeFormat, datetime, time.Local)
 }
 
-// matchStorageToRange identifies video files that overlap with the requested time range (start to end)
+// FormatDateTimeString formats a datetime to a string in UTC with a Z suffix.
+func FormatDateTimeString(datetime time.Time) string {
```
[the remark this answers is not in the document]
hexbabe: `Where is the function to take a string if this format and give me back a time.Time or an error if it fails to parse?`

I guess the closest func is `ExtractDateTimeFromFilename` but it should not be named `FromUTC` because it can also parse localtime file names (files created before our unix timestamp code change)

videostore/utils/utils.go
```diff
@@ -1,6 +1,8 @@
-package videostore
+// Package vsutils provides shared utility functions for video store.
+package vsutils
 
 /*
+#cgo CFLAGS: -I${SRCDIR}
```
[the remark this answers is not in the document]
hexbabe: this is something I added to try to fix the build after moving utils and indexer to their own packages. It seems test and module make targets work even with it removed. Removed.

videostore/indexer/indexer.go
```diff
@@ -0,0 +1,618 @@
...
+}
+
+// Indexer manages metadata for video segments stored on disk.
+type Indexer struct {
+	logger       logging.Logger
+	storagePath  string
+	storageMaxGB int
+	dbPath       string
+	db           *sql.DB
+	dbPtrMutex   sync.RWMutex
```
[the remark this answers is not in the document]
hexbabe: https://github.com/viam-modules/video-store/pull/65/commits/62b0f8ba791cf1d8a53b2921a092ab780ef2962e

videostore/indexer/indexer_test.go
```diff
@@ -0,0 +1,390 @@
...
+	test.That(t, videoRanges.VideoCount, test.ShouldEqual, expectedValidFiles)
+
+	// clean up
+	cancel()
+	<-done
+	err = idx.Close()
+	test.That(t, err, test.ShouldBeNil)
+}
+
+func TestManuallyDeletedFile(t *testing.T) {
```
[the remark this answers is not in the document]
hexbabe: Let me try

videostore/indexer/indexer.go
```diff
@@ -0,0 +1,619 @@
...
+	if ix.db == nil {
+		return errors.New("indexer not setup")
+	}
+
+	start := time.Now()
+
+	// 1. Index new files
+	indexStart := time.Now()
+	if err := ix.indexNewFiles(ctx); err != nil {
+		return err
```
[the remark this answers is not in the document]
hexbabe: I think we should fail because it would be really hard to debug deletions in a bad state

videostore/indexer/indexer.go
```diff
@@ -0,0 +1,650 @@
...
+// Start initializes the database and starts the indexer's background processing loop.
+// It returns an error if the initial database setup fails.
+func (ix *Indexer) Start(ctx context.Context) error {
+	if err := ix.setupDB(ctx); err != nil {
+		return fmt.Errorf("initial database setup failed: %w", err)
+	}
+
+	ix.workers.Add(func(runCtx context.Context) {
+		ix.run(runCtx)
+	})
```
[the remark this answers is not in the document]
hexbabe: aesthetics.

README.md
```diff
@@ -204,6 +204,88 @@ The fetch command retrieves video from local storage, and sends the bytes direct
...
+    "storage_used_gb": 99.98937438707799,
+    "storage_limit_gb": 100,
+    "device_storage_remaining_gb": 697.2153244018555
+  }
+}
+```
+
+> [!NOTE]
+> Swapping `storage_path` will not delete any data, it will simply cause video store to start persisting video data to the new path and preserve
+the old `storage_path` directory with the old videos and the old database, and save new videos in the new path.
```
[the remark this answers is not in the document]
hexbabe: Done

videostore/indexer/indexer.go
```diff
@@ -0,0 +1,650 @@
...
+	ticker := time.NewTicker(refreshInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			ix.logger.Debug("indexer event loop has stopped")
+			return
+		case <-ticker.C:
+			err := ix.setupDB(ctx)
```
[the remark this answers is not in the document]
hexbabe: Done

videostore/videostore.go
```diff
@@ -409,58 +435,6 @@ func (vs *videostore) processFrames(
 	}
 }
 
-// deleter is a go routine that cleans up old clips if storage is full. Runs on interval
-// and deletes the oldest clip until the storage size is below the configured max.
-func (vs *videostore) deleter(ctx context.Context) {
```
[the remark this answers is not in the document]
hexbabe: added back a deleter routine but fallback-ified

videostore/indexer/indexer.go
```diff
@@ -0,0 +1,650 @@
...
+			ix.logger.Debugw("indexed file not found on disk, marking for deletion", "file", indexedFileName)
+			filesMissingFromDisk = append(filesMissingFromDisk, indexedFileName)
+		}
+	}
+
+	if len(filesMissingFromDisk) > 0 {
+		if err := ix.markSegmentsAsDeleted(ctx, filesMissingFromDisk); err != nil {
+			return fmt.Errorf("failed to mark missing disk files as deleted in DB: %w", err)
+		}
+		ix.logger.Debugf("marked %d files missing from disk as deleted in DB", len(filesMissingFromDisk))
```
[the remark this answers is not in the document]
hexbabe: good point it was to DRY, but it would be more intentional to hard-delete

videostore/indexer/indexer.go
```diff
@@ -0,0 +1,650 @@
...
+	if err != nil {
+		ix.logger.Debugw("failed to get video info, unreadable file will not be indexed", "file", fileName, "error", err)
+		return nil
+	}
+	durationMs := info.Duration.Milliseconds()
+
+	query := fmt.Sprintf(
+		"INSERT INTO %s (file_name, start_time_unix, duration_ms, size_bytes, width, height, codec) "+
+			"VALUES (?, ?, ?, ?, ?, ?, ?) "+
+			"ON CONFLICT(file_name) DO NOTHING;",
```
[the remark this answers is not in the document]
hexbabe: result's rows affected would be 0, and error would be 0

why do you ask? should we be logging when this happens?

videostore/videostore.go
```diff
@@ -409,54 +441,90 @@ func (vs *videostore) processFrames(
...
 		}
 	}
 }
 
-func cleanupStorage(storagePath string, maxStorageSizeGB int, logger logging.Logger) error {
-	maxStorageSize := int64(maxStorageSizeGB) * gigabyte
-	currStorageSize, err := getDirectorySize(storagePath)
+func (vs *videostore) cleanupStorage(ctx context.Context) error {
+	if err := ctx.Err(); err != nil {
+		return err
```
nicksanford, line 475: this isn't an error condition

videostore/videostore.go
```diff
@@ -409,54 +441,90 @@ func (vs *videostore) processFrames(
...
+	)
+
+	files, err := vsutils.GetSortedFiles(vs.config.Storage.StoragePath)
 	if err != nil {
 		return err
 	}
 	for _, file := range files {
-		if currStorageSize < maxStorageSize {
+		if err := ctx.Err(); err != nil {
+			return err
```
nicksanford, line 504: this isn't an error condition

hexbabe: changing this return to nil makes the linter unhappy. I added a check in the caller of `cleanupStorage` instead

videostore/videostore.go
```diff
@@ -449,12 +449,12 @@ func (vs *videostore) fallbackDeleter(ctx context.Context) {
 	// This is a best effort to avoid running the fallback deleter immediately on startup to let the indexer
 	// run its cleanup first.
 	select {
-	case <-time.After(fallbackDeleterStartupDelay * time.Minute):
+	case <-time.After(fallbackDeleterStartupDelayMinutes * time.Minute):
```
nicksanford, line 452: given this is the same interval, can't we remove this select statement?

hexbabe: yeah good catch this code would wait 20 min before trying to do work on startup

videostore/indexer.go
```diff
@@ -0,0 +1,451 @@
...
+			}
+		}
+	}
+}
+
+func (ix *indexer) cleanupAndIndex(ctx context.Context) error {
+	// 1. Clean up db according to storage limits
+	deletedFilePaths, err := ix.cleanupDB(ctx)
+	if err != nil {
+		ix.logger.Errorw("failed to cleanup db", "error", err)
```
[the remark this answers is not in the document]
hexbabe: we do. Is this on an older commit?

videostore/indexer/indexer.go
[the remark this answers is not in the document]
nicksanford: nope, I just wanted to make sure it wasn't an error

videostore/indexer/indexer.go
```diff
@@ -0,0 +1,650 @@
...
+// run starts the indexer event loop.
+func (ix *Indexer) run(ctx context.Context) {
+	ix.logger.Debug("starting indexer event loop")
+	ticker := time.NewTicker(refreshInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			ix.logger.Debug("indexer event loop has stopped")
```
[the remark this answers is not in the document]
hexbabe: thanks

videostore/indexer/indexer.go
```diff
@@ -0,0 +1,652 @@
...
+	}
+
+	diskFileEntries, err := ix.getDiskFilesSorted(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get and sort disk files: %w", err)
+	}
+
+	for _, entry := range diskFileEntries {
+		if ctx.Err() != nil {
+			return ctx.Err()
```
nicksanford, line 234: aren't we going to log errors here https://github.com/viam-modules/video-store/pull/65/files#diff-5d306141e7eea456775630360bf7be76123bf98072b91bdc846c4f4ce0fce8c8R174 when the context is cancelled (which is a shutdown condition, not an error condition)?

nicksanford (changes requested):

videostore/indexer/indexer.go
```diff
@@ -0,0 +1,652 @@
...
+// getDiskFilesSorted returns a slice of diskFileEntry for all valid video files on disk,
+// sorted by their extracted timestamp (newest first).
+// Files with unparseable names are logged and skipped.
+func (ix *Indexer) getDiskFilesSorted(ctx context.Context) ([]diskFileEntry, error) {
+	entries, err := os.ReadDir(ix.storagePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read storage directory %s: %w", ix.storagePath, err)
+	}
+
+	sortableFiles := make([]diskFileEntry, 0, len(entries))
```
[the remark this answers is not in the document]
hexbabe: Done

PR #67, opened 2025-05-15
Make DoCommand docs visible in the Viam app by nesting them within the referenced markdownlink section for inline docs

nathan-contino:
Ran `viam module update`, verified that this works, then re-published the current docs until this PR is approved. There's a significant amount of documentation on DoCommand, so it might make sense to increase the visibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,13 +93,13 @@ Additionally, make sure to add your configured data manager service to the `depe
 }
 ```
 
-## DoCommand API
+### DoCommand API
 
-### From/To
+#### From/To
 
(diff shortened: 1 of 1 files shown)


seanavery (approved):
Nice! Thank you. 

I noticed we are just adjusting the header tags, is there some doc on how to correctly format module READMEs for app?

nathan-contino:
Generally speaking, it's best to nest any documentation that you'd like to show up inline and on your registry page _within_ the section referenced from `meta.json`. But beyond that we're still figuring it out!

I don't have merge privileges, so feel free to merge this if it looks good to you. If you'd like me to manually push the changes after that, I'm happy to do so.

PR #75, opened 2025-09-15
RSDK-11896    —     Update video-store to call GetImages on the dep camera

hexbabe:
[RSDK-11896](https://viam.atlassian.net/browse/RSDK-11896)

As part of GetImage -> GetImages, we should update how the frame fetching video-store gets frames to the new API.

TODO:
- [x] Once this is in, make a release and pin viamrtsp's video-store to this version: https://github.com/viam-modules/video-store/releases/tag/v0.0.8-rc8 new version made, PR: https://github.com/viam-modules/viamrtsp/pull/171

Manually test with RealSense
- [x] Fetching color with only color sensor configured:
_Works_
- [x] Fetching color with both color and depth sensor configured. 
_Fails with  12/31/2025, 1:29:42 PM error local-vs-mod.rdk:component:camera/vs     videostore/videostore.go:533 received 2 images from camera rdk:component:camera/oak-d-cam. Multiple image sources is not supported.   log_ts 2025-12-31T18:29:42.438Z_
- [x] Fetching depth with depth, then color configured as `sensors` 
_12/31/2025, 1:30:37 PM warn local-vs-mod.rdk:component:camera/vs     videostore/videostore.go:550 expected image mime type image/jpeg got image/vnd.viam.dep:   log_ts 2025-12-31T18:30:37.082Z_

[RSDK-11896]: https://viam.atlassian.net/browse/RSDK-11896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/videostore/fetch_frames_test.go
+++ b/videostore/fetch_frames_test.go
@@ -124,7 +124,7 @@ func TestFetchFrames(t *testing.T) {
 		test.That(t, len(frameBytes), test.ShouldEqual, 0)
 	})
 
-	t.Run("Succeeds with multiple images (picks first)", func(t *testing.T) {
+	t.Run("Fails with multiple images", func(t *testing.T) {
 		colorImage, err := camera.NamedImageFromBytes(jpegData, "color", rutils.MimeTypeJPEG)
 		test.That(t, err, test.ShouldBeNil)
 		depthImage, err := camera.NamedImageFromBytes(jpegData, "depth", rutils.MimeTypeJPEG)
@@ -153,15 +153,14 @@ func TestFetchFrames(t *testing.T) {
 			Camera:    mockCam,
 		})
 
-		// Wait for frame to be fetched
+		// Wait to see if any frame gets fetched (it shouldn't)
 		time.Sleep(200 * time.Millisecond)
 
-		// Verify frame was stored
+		// Verify no frame was stored (should still be empty)
 		frame := vs.latestFrame.Load()
 		frameBytes, ok := frame.([]byte)
 		test.That(t, ok, test.ShouldBeTrue)
-		test.That(t, len(frameBytes), test.ShouldBeGreaterThan, 0)
-		test.That(t, frameBytes, test.ShouldResemble, jpegData)
+		test.That(t, len(frameBytes), test.ShouldEqual, 0)
 	})
 
 	t.Run("Fails with wrong MIME type", func(t *testing.T) {
--- a/videostore/videostore.go
+++ b/videostore/videostore.go
@@ -529,6 +529,14 @@ func (vs *videostore) fetchFrames(ctx context.Context, framePoller FramePollerCo
 				)
 				time.Sleep(retryIntervalSeconds * time.Second)
 				continue
+			} else if len(namedImages) > 1 {
+				vs.logger.Errorf(
+					"received %d images from camera %s. Multiple image sources is not supported.",
+					len(namedImages),
+					framePoller.Camera.Name(),
+				)
+				time.Sleep(retryIntervalSeconds * time.Second)
+				continue
 			}
 
 			namedImage := namedImages[0]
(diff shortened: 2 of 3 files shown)


videostore/videostore.go
```diff
@@ -502,29 +503,73 @@ func (vs *videostore) Save(_ context.Context, r *SaveRequest) (*SaveResponse, er
...
+			if len(namedImages) != 1 {
+				vs.logger.Errorf(
+					"expected 1 image, received %d from camera: %s, filter source names: [%s], source names: [%s]",
+					len(namedImages),
+					framePoller.Camera.Name(),
+					strings.Join(filterSourceNames, ", "),
+					strings.Join(sourceNames, ", "),
+				)
+				if len(filterSourceNames) == 0 {
+					vs.logger.Warn("no source name was provided. Set a source name in config to receive one image.")
```
oliviamiller: so does a user need to set a source name for videostore to work at all? If thats the case why not error in Validate if they dont put a source name in the config

videostore/fetch_frames_test.go
```diff
@@ -0,0 +1,462 @@
...
+		time.Sleep(200 * time.Millisecond)
+
+		// Verify no frame was stored (should still be empty)
+		frame := vs.latestFrame.Load()
+		frameBytes, ok := frame.([]byte)
+		test.That(t, ok, test.ShouldBeTrue)
+		test.That(t, len(frameBytes), test.ShouldEqual, 0)
+	})
+
+	t.Run("Fails without source_name and multiple images", func(t *testing.T) {
```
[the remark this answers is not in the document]
hexbabe: it would be a lot more expensive, but I think we could just warn and select a default source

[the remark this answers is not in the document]
hexbabe: I want to loop @SebastianMunozP into this decision. What do we think of forcing users to choose a source when its ambiguous vs. auto-selecting one and logging?

videostore/videostore.go
```diff
@@ -502,29 +503,63 @@ func (vs *videostore) Save(_ context.Context, r *SaveRequest) (*SaveResponse, er
...
+					"no images received from camera %s with requested source name [%s], received sources: [%s]",
+					framePoller.Camera.Name(),
+					strings.Join(filterSourceNames, ", "),
+					strings.Join(sourceNames, ", "),
+				)
+				time.Sleep(retryIntervalSeconds * time.Second)
+				continue
+			}
+
+			// Choose first image always (primary image). To select a "secondary" image, use source_name config attr
```
hexbabe: First image in slice will work 99% of the time. Have config attr only as fallback

cmd/create-encode-db/main.go
```diff
@@ -93,18 +93,34 @@ func main() {
...
+		resp, _, err := c.Images(ctx, nil, nil)
 		if err != nil {
 			logger.Info("ignoring error : " + err.Error())
 			continue
 		}
+		if len(resp) == 0 {
+			logger.Info("no images found")
+			continue
+		}
+		if len(resp) != 1 {
```
SebastianMunozP: Could this be called on a camera that supports multiple image types, such as a 3D camera?
In that case, the resp could be more than 1 and this loop would skip it, am I getting this right?

If that's the case, why not just getting a default image?

cmd/create-encode-db/main.go
```diff
@@ -93,18 +93,34 @@ func main() {
...
+			logger.Info("expected 1 image received " + strconv.Itoa(len(resp)))
```
oliviamiller: should this/the mime type log below be error logs?

videostore/videostore.go
oliviamiller: is the order that the images are in the array deterministic? ie the color image is always index 0 and depth image always index 1 between calls?

hexbabe: We also need to get going and fix all the usages of GetImage(s) before January, and I don't want this complexity to block us from having the appropriate development velocity.

cmd/create-encode-db/main.go
hexbabe: fixed

PR #84, opened 2026-01-22
APP-15000 Direct upload via FileUpload API; Save command specify tags and datasets

n0nick:
Introduce an opt-in “direct upload” mode for video-store Save outputs.
When enabled, `save` writes the concatenated clip into a dedicated staging
directory under `storage_path` (avoiding datamanager watched
`upload_path`), and asynchronously uploads the file to Viam App via the
DataSync FileUpload API, with bounded retries and optional
delete-on-success.

Also extend the `save` DoCommand to accept per-request `tags` and
`dataset_ids` (applied only in direct upload mode), and automatically
derive a `videostore-metadata-...` tag from the `metadata` field. Update
module documentation to describe the new configuration and command
fields and their behavior.


n0nick:
Closing this in favor of #85, I might pick up something similar in the future if needed.

PR #87, opened 2026-03-12
RSDK-13042    —      video-store using GetImages continues even if input is invalid repeatedly

hexbabe:
[RSDK-13042](https://viam.atlassian.net/browse/RSDK-13042)

## Bug Summary
**Steps to reproduce :**

Use depth camera e.g. Realsense and hook it up as the underlying camera to a video-store.

Configure the camera to return only a color image at first and store some video, then change it to return only a depth map in its GetImages response e.g. RealSense. 

**Expected behavior :**

Returns depth mime type invalid error, and stops storing video.

**Actual behavior :**

Returns depth mime type invalid error, but continues storing stale color frame as video.

## Manual Tests

Config: 
```
{
  "components": [
    {
      "name": "vs",
      "api": "rdk:component:camera",
      "model": "viam:video:storage",
      "attributes": {
        "camera": "oak-d",
        "sync": "data_manager-1",
        "storage": {
          "size_gb": 1
        }
      },
      "depends_on": [
        "data_manager-1"
      ]
    },
    {
      "name": "oak-d",
      "api": "rdk:component:camera",
      "model": "viam:luxonis:oak-d",
      "attributes": {
        "sensors": [
          "color"
        ]
      }
    }
  ],
  "services": [
    {
      "name": "data_manager-1",
      "api": "rdk:service:data_manager",
      "model": "rdk:builtin:builtin",
      "attributes": {
        "additional_sync_paths": [],
        "sync_interval_mins": 0.1,
        "capture_dir": "",
        "tags": []
      }
    }
  ],
  "modules": [
    {
      "type": "registry",
      "name": "viam_oak",
      "module_id": "viam:oak",
      "version": "latest"
    },
    {
      "type": "local",
      "name": "local-vs",
      "executable_path": "/Users/seanyu/Projects/sean-video-store/bin/darwin-arm64/video-store"
    }
  ]
}
```

Verified video saves well. Change sensors from `["color"]` to `["depth]`. Observed:

`3/12/2026, 1:45:28 PM warn local-vs.rdk:component:camera/vs     videostore/videostore.go:583 expected image mime type image/jpeg got image/vnd.viam.dep:   log_ts 2026-03-12T17:45:28.138Z`

and no more video saved.

After changing back to `["color"]` there is one corrupted video, but the video resource recovers and keeps writing segments.

## PR
1 reviewer

This was not vibe coded sorry for the slop earlier

[RSDK-13042]: https://viam.atlassian.net/browse/RSDK-13042?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

TODO: make an RC, bump viamrtsp, and consume this change there

### DIFF
--- a/videostore/fetch_frames_test.go
+++ b/videostore/fetch_frames_test.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"image"
 	"image/jpeg"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
@@ -20,9 +21,14 @@ import (
 // mockCamera implements camera.Camera for testing fetchFrames behavior.
 type mockCamera struct {
 	camera.Camera
-	name           resource.Name
+	// mu guards fields that may be modified mid-test while fetchFrames reads them
+	// like imagesToReturn. Otherwise this test is unsafe and go test's -race flag
+	// would get really upset.
+	mu             sync.Mutex
 	imagesToReturn []camera.NamedImage
-	returnError    error
+
+	name        resource.Name
+	returnError error
 }
 
 func (m *mockCamera) Name() resource.Name {
@@ -38,6 +44,8 @@ func (m *mockCamera) Images(
 	resource.ResponseMetadata,
 	error,
 ) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
 	if m.returnError != nil {
 		return nil, resource.ResponseMetadata{}, m.returnError
 	}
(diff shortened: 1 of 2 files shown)


videostore/fetch_frames_test.go
```diff
@@ -199,6 +207,72 @@ func TestFetchFrames(t *testing.T) {
...
+			}
+			select {
+			case <-timeout:
+				t.Fatal("timed out waiting for stale frame to be cleared")
+			default:
+				time.Sleep(50 * time.Millisecond)
+			}
+		}
+	})
+
```
seanavery, line 497: Should we add a test case here for valid jpeg mime_type but invalid image bytes? to make sure we clear in that case

seanavery (commented):
Manual tests looking good, but wondering about the case in which the camera GetImages endpoint recovers back to color within a segment boundary. Does the encoder and resulting video file recover?

videostore/videostore.go
```diff
@@ -530,6 +524,12 @@ func (vs *videostore) Save(_ context.Context, r *SaveRequest) (*SaveResponse, er
 	return &SaveResponse{Filename: uploadFileName}, nil
 }
 
+// clearLatestFrame sets the latest frame atomic to a nil byte slice.
+// Useful to remove stale frames when underlying cam is unhealthy.
+func (vs *videostore) clearLatestFrame() {
+	vs.latestFrame.Store([]byte(nil))
```
SebastianMunozP: [nit] We may get a small time win here if we only store a nil byte if its content is not nil already (avoiding the atomic operations). But totally fine the way this is.

hexbabe:
> Manual tests looking good, but wondering about the case in which the camera GetImages endpoint recovers back to color within a segment boundary. Does the encoder and resulting video file recover?

yes I actually tested this but forgot to include in the writeup. Thanks for asking!

edit: Description updated.

videostore/fetch_frames_test.go
hexbabe: yes I will add.

hexbabe: looked into it: there's no explicit path in our code that checks and propagates a sentinel error we own on bad bytes. Want me to add that? Or would you rather assert on err string from underlying lib?

videostore/videostore.go
hexbabe: done

seanavery (commented):
Looks like we currently only test the mime error path with clearLatestFrame. I think it could be worth testing the other paths in which we call `clearLatestFrame`:
- 0 images
- more than 1 image
- Images() error

hexbabe:
> Looks like we currently only test the mime error path with clearLatestFrame. I think it could be worth testing the other paths in which we call clearLatestFrame:
0 images
more than 1 image
Images() error

Great callout for better coverage. Please see if this commit satisfies this end! https://github.com/viam-modules/video-store/pull/87/commits/693942d854238b38121c10a69a5a028534ec83dd

videostore/fetch_frames_test.go
seanavery: Can we just use a mocked NamedImage the errors out on Bytes()?

seanavery (approved):
Looking good, thanks for adding more test coverage. Approving without Bytes error path coverage, fine to merge without it if it's difficult to test.

PR #89, opened 2026-04-29
APP-15996 Route FFmpeg logs to appropriate output streams

katiepeters:
## Summary

  - Routes FFmpeg log output to the appropriate output stream
  based on log level: warnings and above go to `stderr`, while
  info/verbose/debug go to `stdout`
  - Replaces the previous approach of delegating to
  `av_log_default_callback` (which always writes to `stderr`)
  with a custom formatter using `av_log_format_line`
  - Adds a `[FFmpeg <Level>]` prefix to all FFmpeg log lines
  for easier identification
  - Adds log level filtering at the top of the callback to skip
   messages below the current log level threshold

  ## Motivation

  The RDK captures `stderr` output and surfaces it as
  error-level logs. The previous implementation wrote all
  FFmpeg output (including routine info/verbose/debug messages)
   to `stderr`, causing misleading error noise in production.
  This change ensures only genuine warnings and errors appear
  as errors in the RDK log output.

  ## Test plan

  - [ ] Verify FFmpeg info/verbose/debug messages appear on
  `stdout` and not `stderr`
  - [ ] Verify FFmpeg warning and error messages still appear
  on `stderr`
  - [ ] Confirm `[FFmpeg <Level>]` prefix is present on all
  FFmpeg log lines
  - [ ] Check that log level filtering still respects the
  configured `av_log_get_level()`


seanavery:
@katiepeters Thanks for putting this PR up! Let me know if you need help testing.

seanavery:
Ah, sorry we are still using buildjet for CI and it is [stalling again](https://github.com/viam-modules/video-store/actions/runs/25135283067/job/73671958598?pr=89).  Should be able to rebase once #90 is merged.

katiepeters:
> @katiepeters Thanks for putting this PR up! Let me know if you need help testing.

@seanavery could use help testing, thanks!

seanavery:
@katiepeters Added some unit tests and things are looking good. Still need to verify that logs are properly redirected when video-store is used as a library from viamrtsp.

videostore/utils/utils_ffmpeg_log_test.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	rErr, wErr, err := os.Pipe()
+	if err != nil {
+		t.Fatalf("pipe stderr: %v", err)
+	}
+
+	if err := syscall.Dup2(int(wOut.Fd()), syscall.Stdout); err != nil {
+		t.Fatalf("dup2 stdout: %v", err)
+	}
+	if err := syscall.Dup2(int(wErr.Fd()), syscall.Stderr); err != nil {
```
seanavery: Oops. Dup2 does not exist on arm64. Changing

seanavery:
Added a hook in viamrtsp to execute the callback override
[PR](https://github.com/viam-modules/viamrtsp/pull/212)

seanavery:
@katiepeters I think we are getting close to fully testing this : )

Do we want to run a custom build of the viamrtsp module in the problem environment (machine that catalyzed this investigation/fix)?

videostore/utils/utils_ffmpeg_log_cgo.go
```diff
@@ -0,0 +1,55 @@
+package vsutils
+
+/*
```
ale7714, line 3: i think you can merge all of this into utils_ffmpeg_log_test.go or rename this file into `utils_ffmpeg_log_helpers_test.go` to avoid leaking this helpers into the module api

videostore/utils/utils.c
```diff
@@ -82,10 +83,35 @@ int video_store_get_video_info(video_store_video_info *info, // OUT
 }
 
 void video_store_custom_av_log_callback(void *ptr, int level, const char *fmt, va_list vargs) {
-    // Default callback will handle log level filtering
-    av_log_default_callback(ptr, level, fmt, vargs);
```
ale7714, line 86: are we sure we want to stop doing this? i think this is removing some of the repeated log lines suprression

videostore/utils/utils.c
```diff
@@ -82,10 +83,35 @@ int video_store_get_video_info(video_store_video_info *info, // OUT
...
-    // Default callback only prints to stderr
-    fflush(stderr);
+    if (level > av_log_get_level()) {
+        return;
+    }
+    int print_prefix = 1;
+    char line[2048];
```
ale7714: this is truncating the log line by default, no?

ale7714 (commented):
i have a few questions, code looks generally good

videostore/utils/utils_ffmpeg_log_cgo.go
seanavery: Good call on API leak. Made the symbols unexported (test file is in the same package vsutils so it can still access them directly).

AFAIK, CGO isn't allowed in `_test.go` files (Go toolchain restriction), so the file has to live outside the test.

videostore/utils/utils.c
seanavery: [Ref av_log_default_callback implementation](https://github.com/FFmpeg/FFmpeg/blob/master/libavutil/log.c#L380-L439) from the FFmpeg codebase.

The original strategy I developed was to just call this default callback and `fflush` so the logs at least bubbled up to our logger  & app.

From what I can tell, we have to avoid calling this directly as it will route all levels to `stderr` via the `colored_fputs` method.

WRT repeated log line suppression, we are not currently utilizing the `AV_LOG_SKIP_REPEATED` flag to take advantage of this feature. In the future, I am thinking we could vendor this specific part into our callback.

videostore/utils/utils.c
seanavery: ~~This is the standard char length used in the ffmpeg logging codebase so should be good.~~

Looks like ffmpeg logging code uses 1024 max length. [Ref](https://github.com/FFmpeg/FFmpeg/blob/5c44245878e235ae64fe87fb9877644856d33d1d/libavutil/log.c#L51)

We should probably match the `LINE_SZ` define from ffmpeg here.

ale7714: i see, yes agree on matching

seanavery: Done

ale7714 (approved):

seanavery (approved):
LGTM. Tested full-stack usage from viamrtsp

PR #91, opened 2026-05-07
Support cameras with multiple sources

katiepeters:
## Summary
- Add an optional `source_name` config field so a video-store camera can be paired with a multi-source upstream (e.g. RealSense `color`/`depth`).
- When set, it's passed as `filterSourceNames` to `camera.Images()` and used to pick the matching `NamedImage` from the response — defending against upstreams that ignore the filter.
- When unset, behavior is unchanged: the upstream is expected to return exactly one image.

## Test plan
- [x] Manually configure a video-store on top of a RealSense with `source_name: "color"` and confirm clips contain the color stream
- [x] Confirm that single-source cameras still work without additional configuration

### DIFF
--- a/videostore/videostore.go
+++ b/videostore/videostore.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/viam-modules/video-store/videostore/indexer"
 	vsutils "github.com/viam-modules/video-store/videostore/utils"
+	"go.viam.com/rdk/components/camera"
 	"go.viam.com/rdk/logging"
 	"go.viam.com/rdk/resource"
 	"go.viam.com/rdk/services/video"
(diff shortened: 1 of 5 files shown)


videostore/videostore.go
```diff
@@ -538,6 +539,30 @@ func (vs *videostore) clearLatestFrame() {
...
+// camera; not every implementation honors it. Matching by SourceName here
+// keeps a camera that ignores the filter (and returns all its sources) from
+// silently feeding the wrong stream into the video store.
+func pickImage(images []camera.NamedImage, sourceName string) (camera.NamedImage, error) {
+	if sourceName == "" {
+		if len(images) > 1 {
+			return camera.NamedImage{}, fmt.Errorf(
+				"received %d images; set source_name in the config to select one", len(images))
+		}
+		return images[0], nil
```
seanavery, line 556: Should we check `len(images) == 0` before this to future-proof againstc allers of this method panicking here (when source_name is set but list is empty)?

videostore/videostore.go
```diff
@@ -568,18 +597,15 @@ func (vs *videostore) fetchFrames(ctx context.Context, framePoller FramePollerCo
...
-				vs.logger.Errorf(
-					"received %d images from camera %s. Multiple image sources is not supported.",
-					len(namedImages),
-					framePoller.Camera.Name(),
-				)
+			}
+
+			namedImage, err := pickImage(namedImages, framePoller.SourceName)
+			if err != nil {
+				vs.logger.Errorf("camera %s: %v", framePoller.Camera.Name(), err)
```
seanavery, line 604: If the SourceName is wrong, this will probably spam the logs on every frame tick. Seems ok for now, but might want some sort of backoff check eventually.

seanavery (commented):
Nice! Great testing: )
It will be so nice to support video storage form realsense and other depth cams

Make sure we update the readme here with the new source_name attribute.

Looks like you have already manually tested video storage from realsense, just making sure we have also done a live test with single source cameras again.

Heads-up: I think we will also need a follow-up PR on viamrtsp with this new config option

videostore/videostore.go
katiepeters: It would log every `retryIntervalSeconds`

videostore/videostore.go
katiepeters: We could...although it would be an extra check. We check on line 592.

seanavery (approved):

PR #92, opened 2026-05-19
Bump rdk

seanavery:
## Description

RDK dep was super old. Simple bump and test adjustments.

## Testing
- integration tests passing ✅


model/camera/camera.go
```diff
@@ -174,8 +174,8 @@ func (c *component) Close(_ context.Context) error {
 	return nil
 }
 
-func (c *component) Image(_ context.Context, _ string, _ map[string]interface{}) ([]byte, camera.ImageMetadata, error) {
-	return nil, camera.ImageMetadata{}, errors.New("camera.Image not implemented")
+func (c *component) Status(_ context.Context) (map[string]interface{}, error) {
+	return map[string]interface{}{}, nil
```
seanavery, line 178: should we implement Status?

oliviamiller: maybe could return the storage state?

oliviamiller (approved):

jmatth (approved):

model/camera/camera.go
seanavery: Could be a good idea -- i can create a follow-up

